PR #9, opened 2017-09-15
Css overrides

carterbs:


gulpfile.js
```diff
@@ -84,7 +85,10 @@ function watchClients(done) {
 	));
 	done();
 }
-
+function watchSass(done) {
+	watch(path.join(__dirname, './src/components/assets/**/*'), {}, gulp.series(buildSass));
```
NinjaManatee: The ./ is the path.join is unnecessary.

server/FinsembleConfigs.js
```diff
@@ -13,6 +13,7 @@ Finsemble.setDefaultStorage("localStorage");
 
 
 Finsemble.updateBaseUrl(hostname);
+Finsemble.setCSSOverridesPath(StartupConfig[env].clientRoute + '/dist/components/assets/css/finsemble-overrides.css');
```
NinjaManatee: Is this a URL or a path? If it is a path, you should use path.join

PR #12, opened 2017-10-04
install gulp locally rather than using global gulp

NinjaManatee:


rcsharp5:
Not sure this actually uses it. We use gulp 4. You've added gulp 3. Also, the gulp file uses a different version

NinjaManatee:
Hmm, OK, whenever I have installed gulp globally, I just needed npm install
gulp -g. I'll look at the arguments and test it then.


Andy Gardner
*Director, Sales Engineering*
p: 800.821.8147  m: 540.808.5302
w: chartiq.com  e: andy@chartiq.com

PR #26, opened 2018-01-24
5442 clean up linker api

thorsent:
Updated the seed to use the newly refactored Linker Client API.


thorsent:
These changes made their way in when the components were copied wholesale from finsemble-sales-demo

PR #33, opened 2018-02-09
adding a hosted dir to top level

kriswest:
server.js already deals with mounting this at  http://localhost:3375/hosted/ just needs to exist in the seed project. Added a v. brief readme explaining what the dir is for


kriswest:
closing PR so I can create a more extensive one

PR #34, opened 2018-02-12
Kris seed tweaks - improved seed components

kriswest:
Revamped tutorial components:
- sorted functions into order that you uncomment them
- swapped step labels for new numbering
- applied styling to everything based on ChartIQ website (adds weight to CSS, but they never touch that and makes the components feel 'real')
- fixed several slightly incomplete bits (fields with no data, data model is dodgy etc.)


kriswest:
apparently seed changes are supposed to be on another branch for now... closing and resubmitting there

PR #36, opened 2018-02-19
Fixing build to support angular components

kriswest:
Angular components don't currently build themselves properly. This branch includes tweaks to the Gulp file to avoid copying angular component build envs to dist (can take a few mins) and then manually trigger the Angular component builds and output to dist instead.

Took an age to get right, but is essentially a pretty small set of changes. Speeds up the build a lot vs. having an angular component accidentally copied in full by gulp. Also ensures that components generated by CLI work, they did not previously due to issues with the output path (was going into /dist/components/myComponent/dist).

To test:

1. Add the angular CLI globally - DO NOT install it locally or Angular will refuse to generate the source for a nested component project (`You cannot use the new command inside an Angular CLI project.`):
    `npm install -g @angular/cli@latest`
2. Use the *finsemble* cli to generate an angular component:
    `finsemble-cli add component myAngularComponent --component-type=angular`
    Ignore the message:
    ```
    Unable to find "@angular/cli" in devDependencies.

    Please take the following steps to avoid issues:
    "npm install --save-dev @angular/cli@latest"
    ```
    As your generated component project will automatically have that dependency when generated.
3. Build and run with:
    `npm run dev`

The component generated will have a full angular build environment, including a large number of node_modules. It should auto-build itself when you hit `npm run dev` and install the output (no node_modules) in dist.


gulpfile.js
```diff
@@ -120,12 +137,46 @@ function webpackComponents(done) {
...
+
+		var process = function (row) {
+			var compName = row.source.split("/").pop();
+			var cwd = path.join(__dirname, row.source);
+			var command = baseCmd + compName + '/" --outputPath "' + path.join(__dirname, row.source, row["output-directory"]) + '"';
+			
+			// switch to components folder
+			var dir = shell.pwd();
+			shell.cd(cwd);
+			var outputNpm = shell.exec("npm install"); // CLI doesn't install NPM modules, mmake sure this happens
```
rcsharp5: We shouldn't do this here

gulpfile.js
```diff
@@ -6,6 +6,8 @@ var ON_DEATH = require('death')({ debug: false });
 var watch = require("gulp-watch");
 var del = require('del');
 var sass = require('gulp-sass');
+var shell = require('shelljs');
+var eachRow = require('gulp-each-row');
```
rcsharp5: What is the value of this library? Looks like you're just doing a for loop.

gulpfile.js
```diff
@@ -20,13 +22,28 @@ var initialBuildFinished = false;
...
+function buildAngularComponentIgnore() {//Dont copy files built by angular
+	var componentIgnores = [];
+	try {
+		var angularComponents = require('./build/angular-components.json');
+		var arrayLength = angularComponents.length;
+		for (var i=0; i < arrayLength; i++) {
+			componentIgnores.push(path.join('!' + __dirname, angularComponents[i].source, '**'));
+		}
+	} catch (ex) {
+		console.log("Error constructing angular component ignores: " + ex.message + "\n" + ex.stack);
```
NinjaManatee: Shouldn't this be console.error?

gulpfile.js
```diff
@@ -6,6 +6,8 @@ var ON_DEATH = require('death')({ debug: false });
 var watch = require("gulp-watch");
 var del = require('del');
 var sass = require('gulp-sass');
+var shell = require('shelljs');
```
NinjaManatee: How is shelljs exec different from child-process exec?

Why use this rather than the builtins for _dirname and process.chdir?

gulpfile.js
```diff
@@ -120,12 +137,46 @@ function webpackComponents(done) {
...
 function webpackServices(done) {
 	var webpack_config = require('./build/webpack/webpack.services.js')
 	webpack(webpack_config, function (err, stats) {
 		done();
 	})
 }
+
+function angularBuild(done) {
+	try {
+		var baseCmd = 'ng build --base-href "/components/'
```
NinjaManatee: I don't like splitting the quotes like this. Makes it hard to read. Might be better to do something like:

```js
const outputPath = path.join(__dirname, row.source, row["output-directory"]);
const command = `ng build --base-href "/components/${compName}/" -- outputPath "${outputPath}"`;
```

gulpfile.js
NinjaManatee: Is this build required because the Angular components have their own package.json and the CLI doesn't run install? If that is the case, then it probably would be better for the CLI to output a message with the install command, and have this output an error if the install had not been executed.

gulpfile.js
```diff
@@ -120,12 +137,46 @@ function webpackComponents(done) {
...
+			
+			// switch to components folder
+			var dir = shell.pwd();
+			shell.cd(cwd);
+			var outputNpm = shell.exec("npm install"); // CLI doesn't install NPM modules, mmake sure this happens
+			
+			console.log('Executing: ' + command + "\nin directory: " + cwd);
+
+			var output = shell.exec(command);
+			//console.log('Angular output:', output.stdout);
```
NinjaManatee: Should removed commented out code.

gulpfile.js
NinjaManatee: Is it to signal to gulp when the looping is complete?

gulpfile.js
```diff
@@ -120,12 +137,46 @@ function webpackComponents(done) {
...
+			//console.log('Angular stderr:', output.stderr);
+			console.log('Built Angular Component, exit code = ' + output.code);
+			shell.cd(dir);
+		};
+		
+		return gulp
+			.src('./build/angular-components.json')
+			.pipe(eachRow(process));
+	} catch(ex) {
+		console.log("Error constructing angular component ignores: " + ex.message + "\n" + ex.stack);
```
NinjaManatee: Should be console.error

gulpfile.js
```diff
@@ -161,13 +212,15 @@ gulp.task('build', gulp.series(
 	// webpackServices,
 	webpackComponents,
 	// webpackReactComponents,
-	buildSass
+	buildSass,
+	angularBuild
```
NinjaManatee: Is build order important for these tasks? Do the angular components have SASS to build?

kriswest:
image
I'm using the Angular version that the CLI generates (haven't touch the CLI implementation) - which generates a new project using the ng commandline util, so it will produce whatever the latest stable is. E.g. Full version details:
```
Angular CLI: 1.6.8
Node: 8.9.4
OS: win32 x64
Angular: 5.2.5
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cli: 1.6.8
@angular-devkit/build-optimizer: 0.0.42
@angular-devkit/core: 0.0.29
@angular-devkit/schematics: 0.0.52
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.8
@schematics/angular: 0.1.17
typescript: 2.5.3
webpack: 3.10.0
```
The article @rcsharp5 links to is (I believe, but wouldn't warrant, am no Angular dev) somewhat out of date as the ng tool has moved forward quite a bit a wraps the rest of the toolchain.

gulpfile.js
kriswest: Probably shouldn't do it here no... however out of the box the component isn't build-able until this has happened. The CLI should be doing it but isn't - I will work on adding it there. For now I can either add an error check and kill the build with a message, or just run it once for them if node_modules doesn't exist.

gulpfile.js
kriswest: Was allowing me to to loop the JSON angular-components data, without turning each source dir into a stream, I just needed their names. Was also aiming to get multiple builds running in parallel and deal with them completing asynchronously - have since simplified some (the shell commands are now synchronous) so I could probably eliminate it and go back to standard for loop OR we can retain and I can switch process to return async done functions. I doubt this is a huge priority so suspect we should KISS and got to a for loop.

gulpfile.js
kriswest: yep

gulpfile.js
kriswest: I had loads of trouble convincing the built in child-process exec running ng build command in the correct cwd and not inheriting stuff from the seed project build. The only exec I could to get to work (from the handful I tried) was this, which is actually in use in the CLI, where its used to run the project generation commands.

Can re-attempt with the built in versions, but was very happy when did not return the various avalanches of digital vomit I was getting out of other solutions.

gulpfile.js
kriswest: noted

gulpfile.js
kriswest: I suspect trying to to integrate the builds (of seed project with React components and Angular components) would be an arse - they do share many common components (inc. Sass and Webpack) but I'd probably avoid trying to produce a build that deals with both React and Angular together. We may have clients turn up with an existing Angular component, highly customized build process and ultimately a need to separate the builds... 

However, we are generating the Angular project in the CLI (using `ng new`), so if we can get a combined build to work, great. However, we also use whatever the current Angular CLI installed is so we would likely have to test compatibility at different versions and all new major Angular milestones.

Thoughts? I can try going the other way on this, but think it might take a while...

gulpfile.js
kriswest: Angular build compiles its own Sass, our build was trying (and failing) to compile some of the default Sass (I think blowing up on some test). We now ignore Sass in angular components and let ng build handle it and generate its own files...

gulpfile.js
kriswest: Removing the `npm install` entirely - the CLI does do it... It was a nasty interaction between having Angular CLI installed in the finsemble project and in the child project. Angular projects can be components of other build envs, but not another Angular project. I had installed the CLI locally initially (due to Angular warning messages) those message *must* be ignored during component generation - hence will put a big health warning in docs (when I write them).

gulpfile.js
kriswest: done

kriswest:
@NinjaManatee Could you give this another look for me and squash and merge if ok. Have merged it with master changes  which cleaned up the gulpfile some. 

I have a suspicion that there is a neater solution where angular builds into its own dist dir and then our webpack picks up the built files and combines with ours... but this works and doesn't blend the builds which maybe be cleaner in the long run. Thoughts? Either way this should probably get merged now.

build/angular-components.json
```diff
@@ -0,0 +1,6 @@
+[
+    {
+        "source": "./src/components/testAngular",
```
NinjaManatee: Should a test component be committed with the seed project?

gulpfile.js
```diff
@@ -10,6 +10,7 @@
 	const gulp = require("gulp");
 	const sass = require("gulp-sass");
 	const watch = require("gulp-watch");
+	const shell = require('shelljs');
```
NinjaManatee: Double quote (") for consistency. Applies through out file

gulpfile.js
```diff
@@ -65,6 +66,15 @@
 			source.push("!" + path.join(__dirname, componentsToBuild[key].entry));
 		}
 
+		//Dont copy files built by angular
+		try {
+			var angularComponents = require('./build/angular-components.json');
+			var arrayLength = angularComponents.length;
+			for (var i=0; i < arrayLength; i++) {
```
NinjaManatee: Use forEach

gulpfile.js
```diff
@@ -65,6 +66,15 @@
 			source.push("!" + path.join(__dirname, componentsToBuild[key].entry));
 		}
 
+		//Dont copy files built by angular
+		try {
+			var angularComponents = require('./build/angular-components.json');
```
NinjaManatee: Move all requires to the top of the file

gulpfile.js
```diff
@@ -65,6 +66,15 @@
...
+			var arrayLength = angularComponents.length;
+			for (var i=0; i < arrayLength; i++) {
+				source.push(path.join('!' + __dirname, angularComponents[i].source, '**'));
```
NinjaManatee: Add the exclamation point outside of the join since it isn't part of the path.

gulpfile.js
```diff
@@ -65,6 +66,15 @@
...
+			}
+		} catch (ex) {}
```
NinjaManatee: Log error

gulpfile.js
```diff
@@ -87,8 +97,19 @@
 	 * Builds the SASS files for the project. 
 	 */
 	const buildSass = () => {
+		const source = [path.join(srcPath, "components", "**", "*.scss")];
+		
+		//Dont build files built by angular
+		try {
+			var angularComponents = require('./build/angular-components.json');
```
NinjaManatee: Comments from above apply here

gulpfile.js
```diff
@@ -107,6 +128,43 @@
 		});
 	}
 
+	const buildAngular = done => {
+		var processRow = function (row) {
```
NinjaManatee: Use const or let throughout (this should be const).

Use arrow function

gulpfile.js
```diff
@@ -107,6 +128,43 @@
...
+			var compName = row.source.split("/").pop();
```
NinjaManatee: Could you use [path.basename](https://nodejs.org/api/path.html#path_path_basename_path_ext) here?

gulpfile.js
```diff
@@ -107,6 +128,43 @@
...
+			var cwd = path.join(__dirname, row.source);
+			var outputPath = path.join(__dirname, row.source, row["output-directory"]);
+			var command = `ng build --base-href "/components/${compName}/" --outputPath "${outputPath}"`;
+			
+			// switch to components folder
+			var dir = shell.pwd();
+			shell.cd(cwd);
+			console.log('Executing: ' + command + "\nin directory: " + cwd);
```
NinjaManatee: Consistent string delimiters

Use \`\` for token replacement.

gulpfile.js
```diff
@@ -107,6 +128,43 @@
...
+			shell.cd(dir);
+		};
+	
+		try {
+			var angularComponents = require('./build/angular-components.json');
+		} catch (ex) {
+			console.log("No Angular component configuration found");
+		}
+
+		var arrayLength = angularComponents.length;
```
NinjaManatee: Use forEach syntax

src/components/testAngular/README.md
```diff
@@ -0,0 +1,27 @@
+# TestAngular
```
NinjaManatee: This should get moved into a proper tutorial.

src/components/testAngular/e2e/app.e2e-spec.ts
```diff
@@ -0,0 +1,14 @@
+import { AppPage } from './app.po';
```
NinjaManatee: Are these testAngular files automatically generated, or created by you?

gulpfile.js
kriswest: the only expect error was the require failing as file didn't exist. Gone a different way now.

src/components/testAngular/e2e/app.e2e-spec.ts
kriswest: Autogenerated - everything inside the component is autogenerated!

kriswest:
@NinjaManatee I resolved the nasty build times eventually. Webpack was not easy to configure to ignore the angular component's directory(s) when resolving modules, making that stage take ages. It was also being copied in full to dist again. Moving the angular components totheir own dir really simplifies things. 

Could you test this ans then squash and merge both this and finsemble-cli ASAP?

To test, go to your cli checkout of the PR and run
> npm link
then cd back to your checkout of this PR and run:
> npm link @chartiq/finsemble-cli
to link the updated CLI version. 

Then generate a component:
> finsemble-cli add component test-angular --component-type=angular
And that your build runs correctly/quickly:
> npm run dev

Then remove it again:
> finsemble-cli remove component test-angular .\src\angular-components\test-angular
> npm run dev

If all goes smoothly squash and merge please!

build/child_processes/componentBuildProcess.js
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env node
```
NinjaManatee: This file was removed from master and shouldn't be readded.

build/webpack/webpack.files.js
```diff
@@ -5,22 +5,20 @@ const webpack = require("webpack")
...
-	HMRBlacklist = ["testComponent"],//Components to exclude
+
+let enableHMR = true,//Enable Hot Reload
+	HMRBlacklist = ["testComponent"],//Components to explude
 	HMRWhitelist = [],//Only reload these components
 	componentIgnores = [],
 	webpackConfigs = [];//Our list of webpack configs list
 
-if (process.env.NODE_ENV === "production") {// if we are in production turn off hotreload
+if (!process.env.NODE_ENV) {// if we are in production turn off hotreload
```
NinjaManatee: This should be: `if (process.env.NODE_ENV === "production")`

build/webpack/webpack.services.js
```diff
@@ -3,23 +3,22 @@ var fs = require("fs");
...
+
 function getDirectories(srcpath) {
 	return fs.readdirSync(srcpath).filter(function (file) {
 		return fs.statSync(path.join(srcpath, file)).isFile();
 	});
 }
 
 var services = glob_entries(path.join(__dirname, '../../', "/src/services/**/*.js"));
-var clients = glob_entries(path.join(__dirname, '../../', "/src/clients/**/*.js"));
+
```
NinjaManatee: Did you mean to remove your client work?

build/webpack/webpack.services.js
```diff
@@ -3,23 +3,22 @@ var fs = require("fs");
...
 console.log('services', services);
-console.log('clients', clients);
 var entry = {};
 
-//process service files found
 for (var key in services) {
 	var currentPath = services[key];
 	let splitPath = currentPath.split('/services/');
+	console.log(splitPath);
```
NinjaManatee: Should remove debug code

build/webpack/webpack.services.js
```diff
@@ -32,27 +31,11 @@ for (var key in services) {
...
-	}
-	entry[entryKey] = [currentPath];
-}
-
 
 if (Object.keys(entry).length === 0) {
 	return module.exports = null;//If we don't have services there is no need to create an entry json
 };
-//console.log("entry", entry);
+console.log(entry);
```
NinjaManatee: Should remove debug code

gulpfile.js
```diff
@@ -7,7 +7,6 @@
 	const { exec, spawn } = require("child_process");
 	const ON_DEATH = require("death")({ debug: false });
 	const del = require("del");
-	const fs = require("fs");
```
NinjaManatee: You removed my gulpfile-extensions

gulpfile.js
```diff
@@ -29,6 +27,7 @@
 	let angularComponents;
```
NinjaManatee: Take a look at what I did for extensions, to check whether the file exists, and replace it with an empty array if it doesn't. I think that would be cleaner.

gulpfile.js
```diff
@@ -41,42 +40,38 @@
 	// #region Script variables
 	let distPath = path.join(__dirname, "dist");
 	let srcPath = path.join(__dirname, "src");
-	let watchClose;
```
NinjaManatee: Who will watch the watcher? This ended the watch task if the process closes.

gulpfile.js
```diff
@@ -41,42 +40,38 @@
...
 
 	// If you specify environment variables to child_process, it overwrites all environment variables, including
 	// PATH. So, copy based on our existing env variables.
 	const env = process.env;
+	if (!env.PORT) {
```
NinjaManatee: This should be after the check of NODE_ENV. and it should use the hash form. The what it was before the merge.

gulpfile.js
```diff
@@ -41,42 +40,38 @@
...
-				.pipe(sass().on("error", sass.logError))
-				.pipe(gulp.dest(path.join(distPath, "components")));
-		},
-
-		//Dont copy files built by angular
-		if (angularComponents) {
-			angularComponents.forEach(comp => {
-				source.push("!" + path.join(__dirname, comp.source, '**'));
-			});
+	const copyStaticFiles = () => {
```
NinjaManatee: Did you remove the taskMethods object all together?

server/hotreloadmiddleware/client.js
```diff
@@ -7,7 +7,7 @@ var options = {
 	overlay: true,
 	reload: false,
 	sockets: false,
-	log: false,
+	log: true,
```
NinjaManatee: Brad turned this off for performance reasons

build/webpack/webpack.services.js
kriswest: yep - we're not going to support it or user implemented clients so it should go.

kriswest:
@NinjaManatee apologies for the numerous reversions - those should all be fixed and this *should* be ready to merge.

PR #38, opened 2018-02-20
Made it so the "prod" task actually starts a server hosting the Finsemble files

NinjaManatee:
Most of the code for this was already available, and only minor changes were required. These changes include:
- Moved server start to it's own method
- Refactored startServer to use NODE_ENV and PORT passed, if passed in, but use defaults if not
- Only launch Finsemble appication in dev mode.
- Made the "prod" actually start the server


NinjaManatee:
@kriswest @BradleyCarter Would you guys take a look again. I think another change must be necessary because the javascript files are not being built in prod.

PR #39, opened 2018-02-21
Gulpfile cleanup

NinjaManatee:
Simplified and organized the gulpfile which will make life easier for the project reorg

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -7,10 +7,10 @@ const del = require("del");
 const gulp = require("gulp-4.0.build");
 const sass = require("gulp-sass");
 const watch = require("gulp-watch");
+const merge = require("merge-stream");
 const launcher = require("openfin-launcher");
 const path = require("path");
 const webpack = require("webpack");
-const webpack_stream = require("webpack-stream");
 
 // local
 const webpackFilesConfig = require("./build/webpack/webpack.files.js")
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -1,210 +1,182 @@
...
+const launcher = require("openfin-launcher");
+const path = require("path");
+const webpack = require("webpack-stream");
+
+// local
+const webpackFilesConfig = require("./build/webpack/webpack.files.js")
+const webpackServicesConfig = require("./build/webpack/webpack.services.js")
+// #endregion
+
+// #region Constants
```
carterbs: what's region do? or is that just for you?

gulpfile.js
```diff
@@ -1,210 +1,182 @@
...
+	const source = [
+		path.join(srcPath, "components", "**", "*"),
+		"!" + path.join(srcPath, "components", "**", "*.jsx")];
+
+	// Don't copy files that we build
+	for (const key in componentsToBuild) {
+		source.push("!" + path.join(__dirname, componentsToBuild[key].entry));
+	}
+	
+	return merge(
```
carterbs: what's merge? does it merge the streams into one?

gulpfile.js
```diff
@@ -1,210 +1,182 @@
...
-				return;
-			}
-			this.process.exit();
+	return launcher
+		.launchOpenFin({
+			configPath: startupConfig[env].serverConfig
+		})
+		.then(() => {
+			// OpenFin has closed so exit gulpfile
+			process.exit();
```
carterbs: wow. does this solve ctrl+c?????

gulpfile.js
NinjaManatee: It is organizational. In VS code, it folds the region.

gulpfile.js
NinjaManatee: Yes, so the task doesn't complete until all of the streams complete. https://www.npmjs.com/package/merge-stream

carterbs: looks like it. disregard

gulpfile.js
NinjaManatee: I think so. If you close finsemble in dev mode and OF closes, it kills the gulpfile execution.

PR #40, opened 2018-02-27
Hotreload fixes. This will prevent every component from reloading whe…

carterbs:
See comments for explanation.

### DIFF
--- a/server/hotreloadmiddleware/client.js
+++ b/server/hotreloadmiddleware/client.js
@@ -7,7 +7,7 @@ var options = {
 	overlay: true,
 	reload: false,
 	sockets: false,
-	log: true,
+	log: false,
 	warn: true,
 	name: '',
 	autoConnect: true
--- a/server/hotreloadmiddleware/middleware.js
+++ b/server/hotreloadmiddleware/middleware.js
@@ -111,9 +111,12 @@ function createSocketStream(opts) {
 			fn(clients[id]);
 		});
 	}
-	var io = require('socket.io')(opts.server);
+	if (opts.socketServer) {
+		var io = opts.socketServer;
+	} else {
+		var io = require('socket.io')(opts.server);
+	}
 	io.on('connection', function (socket) {
-		console.log("client connected");
 		var id = clientId++;
 		socket.on('disconnect', function () {
 			delete clients[id];
@@ -137,8 +140,6 @@ function createSocketStream(opts) {
 		}
 	};
 }
-
-
 
 function publishStats(action, statsResult, eventStream, log) {
 	// For multi-compiler, stats will be an object with a 'children' array of stats
--- a/server/hotreloadmiddleware/process-update.js
+++ b/server/hotreloadmiddleware/process-update.js
@@ -33,14 +33,8 @@ module.exports = function(hash, moduleMap, options) {
       if (err) return handleError(err);
 
       if(!updatedModules) {
-        if (options.warn) {
-          console.warn("[HMR] Cannot find update (Full reload needed)");
-          console.warn("[HMR] (Probably because of restarting the server)");
-        }
-        performReload();
         return null;
       }
-
       var applyCallback = function(applyErr, renewedModules) {
         if (applyErr) return handleError(applyErr);
 


server/hotreloadmiddleware/process-update.js
```diff
@@ -33,14 +33,8 @@ module.exports = function(hash, moduleMap, options) {
       if (err) return handleError(err);
 
       if(!updatedModules) {
-        if (options.warn) {
```
carterbs: This is the offending block that was causing issues for us. Particularly line 40. Hotreload was built (in general) to operate on a single bundle, on a single webpage. 

This block says "If there are no updated modules, go ahead and reload anyway, something must have gone wrong."

This is great in an SPA. In finsemble, where we have many components all connected to the same hotReload server, this causes bad behavior. 

Example:
I change a file in the toolbar. Toolbar reloads as it should.
App launcher also gets the event that a file was rebuilt. But it gets into this block. None of my app launcher files changed...but there was a change event. Something must be wrong. reload.

Reload, reload, reload, addEventListener, addEventListener, addEventListener - memory leak, crash.

server/hotreloadmiddleware/client.js
```diff
@@ -7,7 +7,7 @@ var options = {
 	overlay: true,
 	reload: false,
 	sockets: false,
-	log: true,
+	log: false,
```
carterbs: cuts down on the cruft in the console.

server/hotreloadmiddleware/middleware.js
```diff
@@ -111,9 +111,12 @@ function createSocketStream(opts) {
 			fn(clients[id]);
 		});
 	}
-	var io = require('socket.io')(opts.server);
+	if (opts.socketServer) {
+		var io = opts.socketServer;
```
carterbs: Pass in the socketServer (if there's one already) into hotReload

kriswest:
This does stop excess rebuilds... however it also prevents reload on changes to CSS, HTML or components.json. The webpack build is triggered but not the reload. Only modificaitons to JS files actually trigger reload now, meaning CSS/components.json tweaks become a pain to review via hotreload...

server/hotreloadmiddleware/middleware.js
```diff
@@ -111,9 +111,12 @@ function createSocketStream(opts) {
...
+	} else {
+		var io = require('socket.io')(opts.server);
```
NinjaManatee: Let’s move all of the requires to the top of the script so we can easily see the dependencies.

server/hotreloadmiddleware/process-update.js
```diff
@@ -33,14 +33,8 @@ module.exports = function(hash, moduleMap, options) {
...
-          console.warn("[HMR] Cannot find update (Full reload needed)");
-          console.warn("[HMR] (Probably because of restarting the server)");
-        }
-        performReload();
```
NinjaManatee: And everything still reloads as appropriate?

NinjaManatee: Kris’ comment seems to suggest that it doesn’t. The whole point of hot reload was to update CSS and HTML on the fly. 

Config changes (component.json) require restart so they are outside the hot reload umbrella, but it would be nice if we could somehow used dynamic config to change it under the hood. Not for this review, just wishful thinking.

kriswest: Apparently they are still rebuilt, but developer has to hit CTRL+R to reload and see them (or whack a change in the JS file)... Webpack only watches the JS (other processes watch the HTML and CSS). However I still think there's a hack somewhere along the line that would make this work (like having the other watch function touch the JS file when they are rebuilt).

NinjaManatee: Again, this is the whole point of the hot reload. Webpack will build the files, but hot reload was added to automatically reload the page to smooth development. If that doesn't work, then this needs to go back to the drawing board.

carterbs:
webpack only rebuilds the files that are part of its bundles. so if you
don't require the CSS into your js file, webpack won't rebuild it. gulp
will, but webpack won't. webpack is driving the reload.

carterbs:
@kriswest please assign either me or andy to this when you're done.

server/hotreloadmiddleware/process-update.js
kriswest: Agree... If possible CSS and HTML changes should cause a component reload - however if that can't be achieved without hotreload hangs then we'll need to improve docs so that devs at least know they need to hit a manual reload.

NinjaManatee:
I spoke with Brad, and it isn't a degradation of performance. There is
still room for improvement, but it continues to work as before with
performance improvements.


Andy Gardner
*Director, Sales Engineering*
p: 800.821.8147  m: 540.808.5302
w: chartiq.com  e: andy@chartiq.com

PR #49, opened 2018-03-14
Only minify in production

NinjaManatee:
Check whether NODE_ENV is production before adding uglify plug-in
Added script to explicitly build in production and development

Dev build is ~60MB (full dist)
Prod build is ~7BM (full dist)


carterbs:
Tested Andy's stuff. Works.

PR #50, opened 2018-03-14
Add no build

carterbs:
Adds a command that allows user to spin up the server and launch openfin without building all of the files. This is useful for testing the results of `build:prod`.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -50,12 +50,12 @@
 	// #endregion
 
 	// #region Task Methods
-	/** 
-	 * Object containing all of the methods used by the gulp tasks. 
+	/**
+	 * Object containing all of the methods used by the gulp tasks.
 	 */
 	const taskMethods = {
-		/** 
-		 * Builds the SASS files for the project. 
+		/**
+		 * Builds the SASS files for the project.
 		 */
 		buildSass: () => {
 			return gulp
@@ -85,7 +85,7 @@
 			return del(distPath, { force: true });
 		},
 
-		/** 
+		/**
 		 * Copies static files to the output directory.
 		 */
 		copyStaticFiles: () => {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -8,6 +8,7 @@
     "prod": "set NODE_ENV=production&&gulp prod --gulpfile=./gulpfile.js",
     "build": "set NODE_ENV=development&&gulp build --gulpfile=./gulpfile.js",
     "build:prod": "set NODE_ENV=production&&gulp build --gulpfile=./gulpfile.js",
+    "dev:no-build": "set NODE_ENV=production&&gulp dev:no-build --gulpfile=./gulpfile.js",
```
NinjaManatee: Called dev:no-build but sets node_env to production

gulpfile.js
```diff
@@ -263,7 +263,8 @@
 		 * Specifies the default task to run if no task is passed in.
 		 */
 		gulp.task("default", gulp.series("dev:run"));
-
```
NinjaManatee: *nitpick* Should have a line between tasks and the method call.

gulpfile.js
```diff
@@ -263,7 +263,8 @@
...
+		/** Launches the application without building the files. For tsting prod builds. */
+		gulp.task("dev:no-build", gulp.series(taskMethods.startServer, taskMethods.launchApplication))
```
NinjaManatee: Task should be called start:no-build or some such. The methods will use the environment specified in NODE_ENV so dev is a misnomer.

gulpfile.js
```diff
@@ -263,7 +263,8 @@
 		 * Specifies the default task to run if no task is passed in.
 		 */
 		gulp.task("default", gulp.series("dev:run"));
-
+		/** Launches the application without building the files. For tsting prod builds. */
```
NinjaManatee: Testing is misspelled.  

*nitpick* new lines in comment are inconsistent with the rest of the file.

PR #51, opened 2018-03-14
Update OF Version (plus note on iFrame settings)

mhmcclung:
-- updated version for next release (aligning with SaleDemo)
-- I was also going to remove "--framestrategy=frames" from run-time arguments but was surprised to see it's already not there.


kriswest:
Does this need reviewers or moving forward somehow - its been sat for a couple of weeks... @htonus @BradleyCarter

PR #54, opened 2018-03-24
Merge loaded extensions with default extensions.

NinjaManatee:
Attempt to load extensions from file, and merge them into default extensions, so all methods are defined. If no extension methods exist, defaults are used.


thorsent:
@carterbs is this PR still necessary now that reorg2 is complete?

carterbs:
Yeah this is still necessary, but the conflicts need to be resolved. This prevents an error if you only define `pre`, but not `post` or `updateServer`

thorsent:
Merged into 2.3.1 cleanup

PR #57, opened 2018-03-30
New Router Config for 2.3.1 (branched off reorg2)

mhmcclung:
With appAssets config to support FinsembleTransport


thorsent:
@htonus I'm thinking that for seed, again to not make any changes. Most customers won't need the FinsembleTransport. We can provide manual instructions on how to set up config if we find it to be necessary for someone.

PR #58, opened 2018-04-03
Add timeout in the server to attempt to capture failed startup. 

carterbs:
Add a catch to the launchOpenfin promise chain to log any errors. Log when openfin closes.


thorsent:
This is encapsulated in my mega-merge

PR #59, opened 2018-04-03
Made the sample auth very simple.

carterbs:


src-built-in/components/authentication/defaultAuthentication.html
```diff
@@ -118,35 +117,96 @@
...
+
+
+
+
+
+
+
+
+
+	/************************************************
```
thorsent: @carterbs We should either document or remove the dynamic config code. I would vote to document it.

PR #61, opened 2018-04-04
Made a task that checks for assim and fails if on mac.

carterbs:


thorsent:
@carterbs can we do this check at runtime instead? I think we get the OS from openfin right? (I'd rather not junk up the seed any more than necessary)

thorsent:
Closed in favor of runtime: https://chartiq.kanbanize.com/ctrl_board/18/cards/6398/details

PR #63, opened 2018-04-09
Cache files for copy

carterbs:
Caches files between sessions so that we don't waste cycles copying files that haven't changed. Watches sass and static files. 

Parallelizes that which can be run in parallel.

Adds some output.


thorsent:
Merged into 2.3.1 PR

PR #65, opened 2018-04-19
JPM bug fix list

thorsent:


configs/openfin/manifest-local.json
```diff
@@ -64,6 +64,10 @@
         "servicesRoot": "http://localhost:3375/finsemble/services",
         "importConfig": [
             "$applicationRoot/configs/application/config.json"
-        ]
+        ],
+        "windowOverrides": {
```
thorsent: @sgd2z can we make it like this? I'm not sure if periods work inside of config items. This layout is much more self-explanatory, and easy to search for:
"monkey-patch":{
   "window.open",
   "window.alert"
}

src-built-in/components/toolbar/components/AutoArrange.jsx
```diff
@@ -9,11 +9,16 @@ export default class AutoArrange extends React.Component {
 		};
 		this.bindCorrectContext();
 		let self = this;
-		FSBL.Clients.RouterClient.subscribe('DockingService.AutoarrangeStatus', function (err, response) {
-			self.setState({
-				isAutoArranged: response.data.isAutoArranged
+		FSBL.Clients.LauncherClient.getMonitorInfo({
```
thorsent: @carterbs getMonitorInfo() defaults to the current window's windowIdentifier. I don't think you need to pass it in here.

src-built-in/components/toolbar/components/AutoArrange.jsx
```diff
@@ -9,11 +9,16 @@ export default class AutoArrange extends React.Component {
...
+			windowIdentifier: FSBL.Clients.LauncherClient.windowIdentifier
+		}, (err, monitorInfo) => {
+			FSBL.Clients.RouterClient.subscribe('DockingService.AutoarrangeStatus', function (err, response) {
```
thorsent: @carterbs there should be a comment here explaining what is happening.

src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
```diff
@@ -146,8 +147,15 @@ Actions = {
 		Actions.blurWindow();
 
 		async.waterfall([
-			Actions.askIfUserWantsToSave,
```
thorsent: @carterbs let's play chess and not checkers. We know that customers are going to ask us to go the other direction (promptOnDirty), so go ahead and put a conditional in place, if(promptOnDirty), so that we have a pathway for both strategies. Later on we will tie that to a preference or config. What we don't want to lose is the logic to handle the prompting and then have to reimplement it later.

src-built-in/components/fileMenu/src/stores/fileMenuStore.js
```diff
@@ -119,25 +119,12 @@ var Actions = {
...
-								resolve(response.choice);
-							});
-						} else {
-							resolve(response.choice);
-						}
-					});
-			} else {
-				resolve();
-			}
+			FSBL.Clients.WorkspaceClient.saveAs({
```
thorsent: @carterbs I think you've gone too far. We don't want to overwrite the actual saved workspace. We just want to make sure the active workspace is saved, and then restored when the sesssion restarts. I think all you need to do is add a "dirty" flag that gets saved with the active workspace.

src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
```diff
@@ -306,43 +314,15 @@ Actions = {
...
 		Actions.blurWindow();
 		let name = data.name;
-		let activeWorkspace = WorkspaceManagementStore.getValue("activeWorkspace");
-		/**
-		 * Actually perform the switch. Happens after we ask the user what they want.
-		 *
-		 */
-		function switchIt() {
+		//Persist active workspace, then switch.
+		FSBL.Clients.WorkspaceClient.saveAs({
```
thorsent: @carterbs put in the promptOnDirty conditional.

thorsent:
Merged through 2.4 seed changes PR

PR #74, opened 2018-05-02
6817 Always on top component (and  6813)

sgd2z:
Adds always on top icons to windowTitleBar and Toolbar.


thorsent:
@jimbunting have you had the opportunity to see this in action? I need to make sure you approve the design.

Fundamentally this code means that every customer is now going to get these new icons even though it was a specific request of JPM. Even though we're not ready to tackle the extensible toolbar and header, we probably should at least incorporate a provision for turning them off other than advising customers to modify the source code.

PR #77, opened 2018-05-08
Dialogs and Styling Overhaul

DanNicolai:
Converted Finsemble to a dark theme and restructured dialogs.  Relevant ticket numbers are:

6808
6780
6320
6294


src-built-in/components/assets/css/CIQ_Seed.css
```diff
@@ -8,14 +8,14 @@ html, body, div, span {
 	margin:0;
 	padding:0;
 	border:0;
-	font-family: 'Open Sans', sans-serif;
+	font-family: 'Roboto', sans-serif;
```
thorsent: @DanNicolai I don't think CIQ_Seed.css is used anymore. It ought to be deleted.

src-built-in/components/assets/css/_dialogs.css
```diff
@@ -80,25 +80,29 @@ body.dialog input::-webkit-outer-spin-button,
...
     outline: none;
 }
 
 body.dialog input {
     -webkit-appearance: none;
-    color: var(--dialog-font-color);
+    color: #FFFFFF;
     text-align: left;
     font-size: 12pt;
+    font-family: Roboto;
```
thorsent: @DanNicolai why are we hard coding font-family and color instead of referencing variables?

src-built-in/components/assets/css/_fontIcons.css
```diff
@@ -1,2 +1,24 @@
-@import './finfont.css';
-@import './_fontIconWhitelabel.css';
+@import './font-finance.css';
+@import './_fontIconWhiteLabel.css';
```
thorsent: @DanNicolai if you put these classes after _fontIconWhiteLabel.css, then there is no way to actually override them.

src-built-in/components/assets/css/_globals.css
```diff
@@ -1,3 +1,3 @@
 @import url('./_variables.css');
-@import url('./_variablesWhitelabel.css');
+@import url('./_variablesWhiteLabel.css');
```
thorsent: @DanNicolai since customers have already started to use Whitelabel (lowercase) let's just live with it. I don't want to have to step JPM through changing their capitalization.

src-built-in/components/assets/css/_variables.css
```diff
@@ -1,142 +1,203 @@
...
+
+    /* Window Frame*/
+    --window-frame-inactive-color: var(--slate);
+    /* --window-frame-active-color: var(--window-frame-inactive-color); */
+
+    /* Window styles*/
+    --scrim-overlay: var(--blue7);
+    --scrim-negative-overlay: var(--red7);
+    
+    /* Complex menus (e.g., Preferences and App Catalog) */
```
thorsent: @DanNicolai we have about one minute in finsemble-time to come up with a better naming convention than "complex-nav" before it's cemented in stone and we can never change it. Perhaps simply, "nav"?

src-built-in/components/searchMenu/searchMenu.css
```diff
@@ -33,7 +29,7 @@ body {
     height: 21px;
     display: flex;
     font-weight: 600;
-    color: var(--primary-accent-color);
+    color: var(--blue1);
```
thorsent: @DanNicolai this looks to be going backward from what we talked about, where we would reference thematic elements (accent-color) rather than direct colors (blue).

src-built-in/components/searchMenu/src/components/ProviderList.jsx
```diff
@@ -31,13 +31,7 @@ export default class ProviderList extends React.Component {
 			<div className="provider">
 				<div className="searchHeader">
 					{providerInfo.provider.displayName}
-
 				</div>
-
```
thorsent: @DanNicolai this looks like bad merge? I think that I added these lines of code into the seed.

PR #79, opened 2018-05-11
6823 auto restart

carterbs:
Dialog, handlers, etc. 

You can check a box in the userPreferences menu and then set a time that the app will restart (daily).


src-built-in/components/userPreferences/src/stores/UserPreferencesStore.jsx
```diff
@@ -39,6 +40,22 @@ var Actions = {
         FSBL.Clients.WorkspaceClient.addWorkspaceTemplateDefinition(newTemplateJSONDefinition, { force: true }, function (err) {
             console.log("addWorkspaceTemplateDefinition result", err);
         });
+	},
+	/**
+	 * By making the value of `finsembke.scheduledRestart` falsy, the application will remove any existing restart timers.
+	 */
+	disableScheduledRestart: () => {
```
thorsent: @carterbs okay for now but we're going to need to circle back with a more generic way of displaying UI for preferences. We can't be asking developers to modify our store every time they want to add a preference.

@jimbunting this is technical debt.

src-built-in/components/userPreferences/src/components/general/ScheduledRestart.jsx
```diff
@@ -0,0 +1,199 @@
+import React from 'react';
```
thorsent: @carterbs given the short deadlines for his code, fine, but this is technical debt. What we want is a generic, reusable way to display a "time" type preference rather than the first of soon to be a million hard coded preferences.

@jimbunting technical debt.

src-built-in/components/fileMenu/src/stores/fileMenuStore.js
```diff
@@ -88,6 +90,33 @@ var Actions = {
 	hideWindow() {
 		FileMenuStore.finWindow.hide();
 	},
+	//If a scheduled restart is invoked, we will open a dialog that gives the user a chance to cancel the restart.
+	listenForScheduledRestart() {
```
thorsent: @carterbs FileMenu isn't the place for this. We've got too much hard coded crap in FileMenu as it is, let's not add to the mess. This code should be in the service.

src-built-in/components/yesNoDialog/src/app.jsx
```diff
@@ -6,6 +6,7 @@ import React from "react";
 import ReactDOM from "react-dom";
 import "../../assets/css/finsemble.css";
 import { FinsembleDialog, FinsembleDialogQuestion, FinsembleDialogButton } from "@chartiq/finsemble-react-controls";
+import Timer from "./timer";
```
thorsent: @carterbs I guess the timer is a useful thing to have in the yes/no dialog. My original plan had been to implement the restart alert as a notification but I don't yet know whether they sufficiently support the timeouts necessary.

@jimbunting what's your thought on whether the restarting alert message should be dialog or notification?

DanNicolai: The advantage of the yes/no dialog is that the user has the ability to cancel the restart. However, it forces the user to stop what they're doing. 

The advantage of the notification would be that it's less obtrusive and allows the user to keep working. If we could add two buttons to the notifications allowing for "restart now" and "cancel", this would seem optimal.

jimbunting: Getting notified that Finsemble is about to restart *should* stop you from working. You don't want to begin something critical, then lose your progress / be unsure of how far you got. I would like a dialog to force the user to make a decision: (cancel/restart now).

https://chartiq.kanbanize.com/ctrl_board/18/cards/6823/details

PR #81, opened 2018-05-16
Notification-CSS (7022)

DanNicolai:
- Relocated notification CSS into a dedicated file.
- Replaced a notification icon with a background image for easier white labeling. 
- Slightly tweaked CSS so it looks good with 3 lines of descriptive text.


thorsent:
@DanNicolai Now that you've got the _notifications.css. Can you take a swing through it and make some of these styles use the _variables? and/or create new variables? The idea would be to make whitelabeling easy. (if this is tricky in light of your plan to semantically improve the variables then let's work through it together).

DanNicolai:
@thorsent I don't see this as being difficult, but before I proceed, I would like to chat and ask a couple questions about white labeling best practices.

DanNicolai:
@thorsent I've gone and created the appropriate variables for notifications and made the references in _notification.css.  I'll be sure to include notifications in my larger styling reorganization that we whiteboarded. 

The path for the .notification-logo background image is still hardcoded. I tried 3 or 4 methods to make this url path into a variable, but nothing ever took. Google searching led me to the conclusion that this is a chrome-specific bug.

PR #88, opened 2018-05-23
Process monitor 7032

rcsharp5:
I updated this to expand all application by default and only show windows there are not in the config list. Also, made this more performant since we were pinging OpenFin every time the process info updated for an application list.

### DIFF
--- a/src-built-in/components/processMonitor/src/components/ExpandedApplication.jsx
+++ b/src-built-in/components/processMonitor/src/components/ExpandedApplication.jsx
@@ -2,50 +2,133 @@
 import { Actions } from "../stores/ProcessMonitorStore";
 import React from "react";
 
-export default function ExpandedApplication(row) {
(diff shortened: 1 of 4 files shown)


src-built-in/components/processMonitor/src/components/ExpandedApplication.jsx
```diff
@@ -2,50 +2,133 @@
...
+                        } else {
+                            newWindowList.push({ window: item, visibility: false })
+                        }
+                    }
+                })
+                self.setState({ childWindows: newWindowList })
+            });
+        })
+    }
+    buildWidowList() {// build out viewable list
```
thorsent: @rcsharp5 you're the "widow maker" ;)

Markusnarillious:
I can add an entire "code R to the bot". It will have tts abilitys and can take voice commands.

PR #96, opened 2018-06-07
Process monitor6963

carterbs:
Revamp of the process monitor, now "For humans".

Test using finsemble branch processMonitor6963

https://github.com/ChartIQ/finsemble/pull/728


configs/application/config.json
```diff
@@ -51,14 +52,21 @@
 					"Toolbar",
 					"App Launcher",
 					"Workspace Management Menu",
-					"Finsemble Linker Window",
+					"linkerWindow",
```
carterbs: Moved all of these components out of the default agent so they won't show up in "simple view", which hides the system components process.

configs/application/presentationComponents.json
```diff
@@ -217,7 +217,7 @@
 				}
 			},
 			"component": {
-				"spawnOnStartup": true,
+				"spawnOnStartup": false,
```
carterbs: These two were components we weren't using

PR #97, opened 2018-06-10
Inject floated 5966

thorsent:
This adds an absolutely positioned drag handle to the window header. This will replace all of the webkit-app-region: drag stuff that we baked into the header.

The reason for this is because chromium has a bug where the fixed drag region becomes unreachable when the window is scrolled. This absolutely positioned drag handle is a way to deal with that problem.

Status:
- Need to find a way to test this. Currently welcome component works fine with scroll (it didn't in the past?) and I can't get adhoc components working (known site foxnews.com has the scroll problem but I can't get it to show)
- Currently all draggable stuff is still working and a thin red bar appears where the drag handle is positioned. In production we'll make it full height and get rid of the background color.
- Need to hide/show the drag header during drop operations probably?


carterbs:
@thorsent about to review this. heads up that I can't change the base to develop without 150 commits showing up. Don't know what's wrong there.

PR #98, opened 2018-06-11
Whitelabeling - #7172

DanNicolai:
Runs linked to Finsemble Master.

- Variables were put in the _theme.css file and replacements were made across seed. 
- Overrides for those variables can be placed in _themeWhitelabel.css, or _variablesWhitelabel.css as a fallback.
- Tested overriding fonts, icons and classes with override files, everything appears to be in working order.


carterbs:
Same PR as tabs and styles.

PR #107, opened 2018-06-22
increase URL length allowed in adHocComponents

kriswest:
Increase max input length on adHocComponent form (only allowed 40 char URLs previously)


thorsent:
@kriswest @NinjaManatee fixed in "develop" branch. We've eliminated the character limit.

PR #111, opened 2018-06-26
Save active workspace when saveAs/new/export workspace

carterbs:
On saveAs, new Workspace, and export, check to see if autosave is enabled. If so, save the workspace before performing the action.

We should consider moving this functionality into the workspaceClient API. For now, the responsibility lies with the UI.


DanNicolai:
Tested this and appears to work initially. Further testing revealed that there's windows persisting through workspaces, but I'm 80% sure that's a separate issue.

PR #115, opened 2018-06-27
Show search menu below launching component

NinjaManatee:
If the height of the toolbar is changed to have multiple rows, and search is placed on the second row, the search menu will still appear below the first row. This change uses the bounds of the launching element to set the position of the searchMenu.


src-built-in/components/toolbar/stores/searchStore.js
```diff
@@ -59,17 +59,19 @@ var Actions = {
...
+				// 	position: 'relative',
+				// 	left: target.getBoundingClientRect().left,
+				// 	forceOntoMonitor: true,
+				// 	top: 'adjacent',
+				// 	spawnIfNotFound: true
+				// };
 
 				var bounds = document.getElementById("inputContainer").getBoundingClientRect();
-				menuWindow.showAt(window.screenX + bounds.left, target.offsetHeight + window.screenY, null, function (err, data) {
+				menuWindow.showAt(window.screenX + bounds.left, bounds.bottom + window.screenY, null, function (err, data) {
```
thorsent: @NinjaManatee I think we should be using LauncherClient.showWindow(top:"adjacent")

We should also be using WindowClient.isMouseOverDOMElement() instead of mouseInBounds.

Ugh, just kills me when we don't use our own convenience functions...

carterbs: @NinjaManatee if inputContainer isn't there for some reason you're going to get a nasty javascript exception.

src-built-in/components/toolbar/stores/searchStore.js
```diff
@@ -59,17 +59,19 @@ var Actions = {
...
-				let params = {
-					monitor: 'mine',
-					position: 'relative',
-					left: target.getBoundingClientRect().left,
-					forceOntoMonitor: true,
-					top: 'adjacent',
-					spawnIfNotFound: true
-				};
+
+				// TODO: Should we keep this to use later, or remove it?
```
carterbs: remove as much unused code as possible

src-built-in/components/toolbar/stores/searchStore.js
carterbs: @thorsent re: showWindow, it was auto-focusing causing issues with the search menu's animations and stuff. Ryan used showAt to circumvent that problem.

NinjaManatee: @thorsent @carterbs I think I've addressed your comments. I don't want to change to using showWindow because of the issue Brad mentioned. I've added a comment for future clarity.

PR #120, opened 2018-07-02
Recall zoom 7837

thorsent:


carterbs:
@thorsent Does this need review?

thorsent:
@carterbs I'm going to add remembering scroll position as well as zoom, but zoom is done and I wouldn't mind some eyes on it.

thorsent:
Based on feedback from JPM, they need zoom much more than scroll, so pushing this through review as just zoom.

PR #122, opened 2018-07-05
Linker tester

NinjaManatee:
Simple linker tester component to include with the seed project. Component can be linked to other components, specify the data type and data to be sent over the linker. 
image

### DIFF
--- a/configs/application/components.json
+++ b/configs/application/components.json
@@ -169,7 +169,7 @@
         },
         "Linker Tester": {
             "window": {
-                "url": "$applicationRoot/components//linkerTester/linkerTester.html",
+                "url": "$applicationRoot/components/linkerTester/linkerTester.html",
                 "frame": false,
                 "resizable": true,
                 "autoShow": true,
--- a/configs/application/components.json
+++ b/configs/application/components.json
@@ -167,7 +167,7 @@
                 }
             }
         },
-        "linkerTester": {
+        "Linker Tester": {
             "window": {
                 "url": "$applicationRoot/components//linkerTester/linkerTester.html",
                 "frame": false,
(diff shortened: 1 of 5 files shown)


configs/application/components.json
```diff
@@ -1,171 +1,208 @@
...
+                    "Window Manager": {
+                        "FSBLHeader": false,
+                        "persistWindowState": false
+                    }
+                }
+            }
+        },
+        "Linker Tester": {
+            "window": {
+                "url": "$applicationRoot/components//linkerTester/linkerTester.html",
```
carterbs: double slash

src-built-in/components/linkerTester/linkerTester.js
```diff
@@ -0,0 +1,94 @@
...
+			titleDiv.appendChild(document.createTextNode(title));
+
+			const dataBody = document.createElement("pre");
+			dataBody.innerHTML = data;
+
+			received.appendChild(titleDiv);
+			received.appendChild(dataBody);
+		};
+
+		FSBL.addEventListener("onReady", () => {
```
carterbs: For the sake of readability I'd request that you take these callbacks and put them into named functions. We're making a hard push for readability on the other side, so I think we should push for it here as well.

src-built-in/components/linkerTester/linkerTester.js
```diff
@@ -1,94 +1,150 @@
...
+		currentDataType = dataType.value;
+	};
+
+	/**
+	 * Handles the callback for getComponentState
+	 * @param {*} err 
+	 * @param {*} state 
+	 */
+	const componentStateHandler = (err, state) => {
+		return;
```
carterbs: ????

carterbs: why return at the top?

NinjaManatee: Debugging code. Fixed.

thorsent:
The LinkerTester was moved into the finsemble-dev repo

PR #123, opened 2018-07-06
7483 Add tabbing and tiling configs to be able to disable

clausjs:


configs/application/config.json
```diff
@@ -16,6 +16,12 @@
 				"title": "App Catalog",
 				"component": "App Catalog"
 			}
+		},
+		"docking": {
+			"enabled": true,
+			"groupTileBuffer": 30,
+			"headerHeight": 32,
+			"MINIMUM_WIDTH": 175
```
carterbs: @clausjs docking config is now in two places. It should only live in the servicesConfig top-level property. Can't test without knowing which place to modify. Let me know when this is fixed.

PR #124, opened 2018-07-06
Added a user preference for switching toolbar monitor.

thorsent:
Adds a user preference for selecting which monitor the toolbar appears on. User must restart for it to take effect (and they must restart with the button to ensure that localStorage is available).

### DIFF
--- a/src-built-in/components/userPreferences/src/components/general/Toolbar.jsx
+++ b/src-built-in/components/userPreferences/src/components/general/Toolbar.jsx
@@ -11,7 +11,7 @@ export default class Toolbar extends React.Component {
 	}
 
     /**
-     * Sets the meridiem for the time (AM or PM). Afterwards, this calls setHour to handle any necessary conversions to military time.
+     * Sets the monitor by calling the setPreferences API. This will override the config for the component.
      * @param {event} e
      */
 	setMonitor(e) {
--- a/src-built-in/components/userPreferences/src/components/general/Toolbar.jsx
+++ b/src-built-in/components/userPreferences/src/components/general/Toolbar.jsx
@@ -1,0 +1,63 @@
+import React from 'react';
+
+export default class Toolbar extends React.Component {
+	constructor(props) {
+		super(props);
+		this.state = {
+			monitor: "all"
+		};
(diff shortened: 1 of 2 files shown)


src-built-in/components/userPreferences/src/components/general/Toolbar.jsx
```diff
@@ -0,0 +1,63 @@
...
+		super(props);
+		this.state = {
+			monitor: "all"
+		};
+		this.setMonitor = this.setMonitor.bind(this);
+		this.restartApplication = this.restartApplication.bind(this);
+	}
+
+    /**
+     * Sets the meridiem for the time (AM or PM). Afterwards, this calls setHour to handle any necessary conversions to military time.
```
carterbs: docs are wrong

src-built-in/components/userPreferences/src/components/general/Toolbar.jsx
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	componentWillUnmount() {
+	}
+
+	render() {
+		return <div className="complex-menu-content-row">
+			<div>
+				<span>Toolbar Monitor (requires restart)</span>
+				<select style={{ margin: "0px 10px" }} onChange={this.setMonitor} value={this.state.monitor}>
```
carterbs: This should get all monitors from the launcher and just 

```
monitors.map((monitor, i)=>{
    return <option value={i}>{i}</option>
});
```

Don't know if you did this just for protyping, but it should definitely get the monitors from the launcher. Think it's getMonitorInfoAll

thorsent: The reason I didn't do that is because you may move to another system with fewer monitors. The actual value that is set for the monitor will no longer match, it won't be an option. You then get start getting into writing a lot of code to deal with a minor issue. Considering that we plan on moving to a "perfect toolbar", and that this code is essentially temporary, the hard coded values seemed reasonable.

src-built-in/components/userPreferences/src/components/general/Toolbar.jsx
thorsent: Resolved

PR #125, opened 2018-07-10
7966 Hack the scrollbar to appear underneath the window title bar.

thorsent:
Scrollbars now appear underneath window title bar (via a css hack built into the windowTitleBar,jsx) *except* when Window Manager.hackScrollbar=false


carterbs:
image
bad interaction with the zoom PR.

thorsent:
@kris, I multiple body zoom by 10%. I then divide the header zoom by the same amount to keep them proportional.

thorsent:
There's possibly a more clever way to do this, perhaps zooming body:not(#FSBLHeader)...

PR #126, opened 2018-07-11
7295 Adds showWarningForLogStates prop to config for logger service

clausjs:
This will allow the user to easily change the log states which will throw a warning

Finsemble PR: https://github.com/ChartIQ/finsemble/pull/802

### DIFF
--- a/configs/application/config.json
+++ b/configs/application/config.json
@@ -16,6 +16,9 @@
 				"title": "App Catalog",
 				"component": "App Catalog"
 			}
+		},
+		"logger": {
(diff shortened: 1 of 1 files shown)


carterbs:
@thorsent may want to consider defaults here. I'd err on the side of debug/verbose, but mike wanted info in there too. Don't care too much but we should put some thought into it.

carterbs:
@clausjs can this be deleted? Seems unimportant.

clausjs:
@BradleyCarter seems like it, the important part was merged into Finsemble

PR #128, opened 2018-07-11
7837 Implement zoom with browser hotkeys because customers may not ha…

thorsent:
Reimplement zoom using browser hotkeys, because not all customers have assimilation


carterbs:
@thorsent did you forget to add the hackScrollbar functionality? Zoom looks good.

PR #129, opened 2018-07-15
Add globals.d.ts that sucks in type definition for FSBL

carterbs:


kriswest:
:+1 for itellisense in seed, often requested! Particularly useful when docs
are not being updated as at present!

carterbs:
@mesner not sure what to do with this. Looks like I reached out in Jan for a decision. Please decide whether to keep this open or to close it.

mesner:
I thought this was already in as well.  I'd say merge.  It is one line.

carterbs:
@mesner looked into it, and the mechanism that does this no longer works. It was to convert js to ts declaration files. all of our clients are TS now. so we'll need to go back to the drawing board to extract type definitions from our TS files.

PR #140, opened 2018-07-30
add in a check to see if the companion window needs to show itself on…

rcsharp5:
https://chartiq.kanbanize.com/ctrl_board/18/cards/8458/details


rcsharp5:
Companion window now shows after it is in the correct location. Must use finsemble PR https://github.com/ChartIQ/finsemble/pull/838

PR #145, opened 2018-08-10
8383 Webpack 4 plus Repo includes

thorsent:
Adds "include" capability to the build process. This allows other repos to be "overlaid" on top of finsemble-seed. For instance, we want to build a bunch of development tools (Andy built one called the linkerTester) but we don't want to pollute the finsemble-seed project with our internal stuff. "includes" solve this problem.

`npx gulp dev --include:../finsemble-dev`

This command is the same as npm run dev but it will "include" the finsemble-dev directory in the build process. Any components in finsemble-dev will be built. Any configs will be copied into the dist. Any assets will be copied. It is as if you've applied an overlay to the finsemble-seed project.

Yeah, typing .. is a nuisance so you can leave that out for convenience:

`npx gulp dev --include:finsemble-dev`

Note: Since the build process can now take webpack entry points directly from inside components (finsemble.webpack.json) there's no need to modify that pesky build directory.

There's a new repo "finsemble-dev" which can be cloned to try this out. It contains Andy's linkerTester component.

**Notes**

This project required migrating to webpack 4 and babel 7. package.json has changed. You'll need to run `npm install` in order to test.

webpack watch is not working too well with webpack 4 (it's trying to load babel presets from the openfin directory which makes me think that hmr is somehow automatically enabled. It'll take some more fiddling to get that fixed).


thorsent:
I like the idea of a link to dev tools from the central logger!! @kriswest 

The problem that we haven't solved yet with the hosted finsemble is how to do development on core finsemble. This often involves making changes to core, running on local machine, and making changes to presentation components simultaneously. Without this PR (8383), the only way to effectively code core Finsemble is by hacking up the seed.

What would solve this problem would be:
npm run dev manifest=file://blah.json

That could start the local dev server and still pull the manifest from an alternate location, allowing for components and finsemble core to be distributed as needed.

carterbs:
@thorsent Can you make a decision on whether this is necessary any more?

thorsent:
@carterbs considering the move towards dynamic assembly, I don't think the includes are going to be necessary anymore.

There was some effort in here related to getting on webpack 4 which I guess we can resurface if we decide to tackle that.

PR #151, opened 2018-08-14
Authentication component clean up

NinjaManatee:
Split CSS and JS from HTML file
JS is essentially unchanged, so will still need clean up.


configs/application/config.json
```diff
@@ -10,6 +10,13 @@
 			"LocalStorageAdapter": "$applicationRoot/adapters/LocalStorageAdapter.js"
 		}
 	},
+	"isAuthEnabled": true,
```
thorsent: @NinjaManatee isAuthEnabled has been deprecated. Simply putting the startup adapter in place is enough to start up authentication.

But, do you want to modify seed so that authentication is turned on by default?

NinjaManatee: I didn't mean to turn on authentication by default. 

Part of me wants to enable it because it is more in line with how it will likely be used, but may confuse some developers new to Finsemble.

PR #156, opened 2018-09-06
7829 Index db adapter

kriswest:
The IndexDB adapter appears to work well and resolves issues caused by LocalStorage during workspace  saves, where edits to the workspace just before a close or workspace switch can be lost - likely due to localstorage not flushing before the switch.

The storage adapter appears (no formal measurement) at certain operations than localstorage and has explicit callbacks/promises guaranteeing operations have finished before returning. 

Citadel would like access to the adapter through the seed project ASAP as it should resolve most of their workspace issues immediately. Hence, if possible can we expedite a merge.

 We should also consider switching the seed's default storage to IndexDB as that may resolve issues with OF9 (though not with the chart) as well as making workspace demos much more reliable. If not comfortable with changing the default to IndexDB in master then we should at least include the adapater and build it, so that clients can use config to switch to it.


src-built-in/adapters/indexDBAdapter.js
```diff
@@ -0,0 +1,188 @@
...
+// var Logger = require("@chartiq/finsemble").Clients.Logger;
+// var Dexie = require('dexie');
+
+import finsemble from "@chartiq/finsemble";
+import Dexie from 'dexie';
+var BaseStorage = finsemble.models.baseStorage;
+var Logger = finsemble.Clients.Logger;
+
+
+// var BaseStorage = require("@chartiq/finsemble").models.baseStorage;
```
NinjaManatee: Should remove commented out code

src-built-in/adapters/indexDBAdapter.js
```diff
@@ -0,0 +1,188 @@
...
+		
+		Dexie.spawn(function*() {
+			yield db.fsbl.put({key: combinedKey, value: params.value});
+		}).then(function() {
+		   // spawn() returns a promise that completes when all is done.
+		   return cb(null, { status: "success" });
+		}).catch(function(e) {
+		   console.error("Failed: " + e);
+		   //TODO: can't help think this should be (but localstorage adapter doesn't return errors...)
+		   //return cb({ status: "failed" }, null);
```
NinjaManatee: Do the return correctly and take the TODO out.

src-built-in/adapters/indexDBAdapter.js
```diff
@@ -0,0 +1,188 @@
...
+		Dexie.spawn(function*() {
+			let val = yield db.fsbl
+				.where('key').equals(combinedKey).first();
+			Logger.system.debug("Storage.getItem for key=" + combinedKey + " raw val=" + val);
+			console.debug("Storage.getItem for key=" + combinedKey + " val=", val);
+			let data = {};
+			if (val && val.value) {
+			 	data = val.value;
+			 }
+			// Logger.system.debug("Storage.getItem for key=" + combinedKey + " with data=" + data);
```
NinjaManatee: You should either uncomment this or remove it.

src-built-in/components/authentication/defaultAuthentication.html
```diff
@@ -175,7 +175,7 @@
 		FSBL.Clients.WindowClient.close({ removeFromWorkspace: false, closeWindow: true });
 	}
 
-	//CLose app when the X is clicked.
+	//Close app when the X is clicked.
```
NinjaManatee: You should take this out of this PR

src-built-in/adapters/indexDBAdapter.js
```diff
@@ -0,0 +1,188 @@
...
+
+Logger.system.debug("IndexDBAdapter", Dexie);
+console.debug("IndexDBAdapter", Dexie);
+
+var IndexDBAdapter = function (uuid) {
+	BaseStorage.call(this, arguments);
+
+	var db = new Dexie('finsemble');
+	db.version(1).stores({
+		fsbl: `key` //only index the storage keys
```
NinjaManatee: Why the `?

src-built-in/adapters/indexDBAdapter.js
```diff
@@ -0,0 +1,188 @@
...
+	 * @param {string} params.topic A topic under which the data should be stored.
+	 * @param {string} params.key The key whose value is being set.
+	 * @param {any} params.value The value being saved.
+	 * @param {function} cb callback to be invoked upon save completion
+	 */
+	this.save = function (params, cb) {
+		Logger.system.debug("saving", params);
+		const combinedKey = this.getCombinedKey(this, params);
+		
+		Dexie.spawn(function*() {
```
NinjaManatee: Why does spawn need a generator?

src-built-in/adapters/indexDBAdapter.js
```diff
@@ -0,0 +1,188 @@
...
+		let preface = self.baseName + ":" + self.userName + ":" + params.topic + ":";
+		if ("keyPrefix" in params) {
+			preface = preface + params.keyPrefix;
+		}
+		return preface;
+	};
+
+	/**
+	 * Returns all keys stored in localstorage.
+	 * @param {*} params
```
NinjaManatee: Documentation?

src-built-in/adapters/indexDBAdapter.js
```diff
@@ -0,0 +1,188 @@
...
+			//TODO: can't help think this should be (but localstorage adapter doesn't return errors...)
+			//return cb({ status: "failed" }, null);
+		});
+	};
+
+	/**
+	 * This method should be used very, very judiciously. It's essentially a method designed to wipe the database for a particular user.
+	 */
+	this.clearCache = function (params, cb) {
+		const preface = self.baseName + ":" + self.userName;
```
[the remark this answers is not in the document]
kriswest: In the other methods keyPreface includes a topic, where this the userPreface I suppose.

src-built-in/adapters/indexDBAdapter.js
kriswest: The generator really changes how the function is executed - inside the generator you can use the yield keyword to wait for an async method (promise) to return as if it were synchronous. The generator itself returns a promise so the whole block is async, but you can write nice simple synchronous looking code inside.

All the db.fsbl calls are promise based - see http://dexie.org/docs/Simplify-with-yield

configs/application/config.json
```diff
@@ -1,19 +1,21 @@
 {
 	"comment": "Top-level application config for development, adding on top of the core config",
+	"defaultStorage": "IndexDBAdapter",
```
NinjaManatee: I'm not sure whether we want this to be default.

src-built-in/adapters/indexDBAdapter.js
```diff
@@ -0,0 +1,199 @@
...
+
+const IndexDBAdapter = function () {
+	BaseStorage.call(this, arguments);
+
+	const db = new dexie("finsemble");
+	db.version(1).stores({
+		fsbl: "key" //only index the storage keys
+	});
+
+	Logger.system.debug("IndexDBAdapter init");
```
NinjaManatee: Is it necessary to log to both the Logger and the console?

kriswest:
Ready for you to review @thorsent

thorsent:
I don't see any reason not to make this the default once we improve the adapter interface and eliminate the dependency. indexedDB seems to have no downside.

kriswest:
@thorsent I get why we should avoid the dependency. I've rebuilt this with the base IndexedDB API, could you take another look? I'm keen to get this on to Citadel to try out see if it resolves their various localstorage issues.

Would it be possible to get this merged before updating the storage adapter interface (which I imagine will take a while) - if you'd rather not switch the default, we could just include it and advise clients on switching...

src-built-in/adapters/indexedDBAdapter.js
```diff
@@ -0,0 +1,259 @@
...
+
+const BaseStorage = finsemble.models.baseStorage;
+const Logger = finsemble.Clients.Logger;
+
+// Because calls to this storage adapter will likely come from many different windows, we will log successes and 
+// failures in the central logger.
+Logger.start();
+
+const IndexedDBAdapter = function () {
+	BaseStorage.call(this, arguments);
```
NinjaManatee: @thorsent I tried to change this to an ES6 class, but I got error when this method was called as:
```super(arguments);```
If you rework this interface, we should keep an eye on being able to support ES6 classes.

File: src-built-in/adapters/indexedDBAdapter.js [hunk 12]
```diff
@@ -0,0 +1,259 @@
...
+			return cb(event, { status: "failed" });
+		};
+		request.onsuccess = function(event) {
+			cb();
+		};
+	};
+};
+
+IndexedDBAdapter.prototype = new BaseStorage();
+new IndexedDBAdapter("IndexedDBAdapter");
```
NinjaManatee: @thorsent The fact that this script instantiates the IndexDBAdapter and does nothing seems very weird to me. I would have thought it would export the instance like the services does, or, even better, return the constructor so it can be instantiated by Finsemble. Either way, I think this could be made clearer during refactoring.

src-built-in/adapters/indexedDBAdapter.js
```diff
@@ -0,0 +1,259 @@
...
+	/**
+	 * Wipes the storage container.
+	 * @param {function} cb
+	 */
+	this.empty = (cb) => {
+		Logger.system.debug("IndexedDBAdapter.empty");
+		console.debug("IndexedDBAdapter.empty");
+		var objectStore = db.transaction(["fsbl"], "readwrite").objectStore("fsbl");
+		var request = objectStore.clear();
+		request.onerror = function(event) {
```
NinjaManatee: I think these error functions are essentially the same through out the file, so you might be able to create an error handler function to simplify the code.

src-built-in/adapters/indexedDBAdapter.js
thorsent: Agreed.

thorsent:
@kriswest would love to see a comment around what IDBKeyRange() and successor() to save some future developer the time to figure out what it is.

kriswest:
Merged Andy's cleanup and added requested comments. Re-tested and appears to work fine. 
@thorsent should be ready for review and merge.

PR #157, opened 2018-09-06
Index db adapter cleanup

NinjaManatee:
I think this resolves all of the issues I raised from my review. Sadly I wasn't able to make it an ES6 class because of a weird error where `setBaseName` was undefined. Should probably figure out why that is, and fix it in BaseStorage.


kriswest:
Looks fine to me - although I'm having trouble with the clearCache method, which is not clearing said cache

PR #169, opened 2018-09-24
#9212 fix for search in catalog

rcsharp5:
We hardcoded a c into the catalog store. I just removed that. This also requires a fix on the FInsemble side.


rcsharp5:
https://chartiq.kanbanize.com/ctrl_board/18/cards/9212/details

PR #170, opened 2018-09-24
9324 IndexedDBAdapter Queuing

thorsent:
Implemented simple queuing for indexedDBAdapter to capture storage requests prior to initialization, and service them immediately when initialization is complete.

### DIFF
--- a/src-built-in/adapters/indexedDBAdapter.js
+++ b/src-built-in/adapters/indexedDBAdapter.js
@@ -72,6 +72,11 @@ const IndexedDBAdapter = function () {
 	 * The IndexedDB instance.
 	 */
 	this.db;
+
+	/**
+	 * Array of commands received before the IndexedDB is initialized. These commands are executed once the connection
+	 * has been established.
+	 */
 	this.queue = [];
 
 	Logger.system.debug("IndexedDBAdapter init");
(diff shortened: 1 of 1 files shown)


src-built-in/adapters/indexedDBAdapter.js
```diff
@@ -134,6 +137,15 @@ const IndexedDBAdapter = function () {
 		return preface;
 	}
 
+	this.releaseQueue = () => {
+		if (this.queue.length) {
+			Logger.system.debug("IndexedDBAdapter.releaseQueue");
```
NinjaManatee: It might be better to make this command always log with:
```Logger.system.debug(`IndexedDBAdapter.releaseQueue: ${this.queue.length} commands`);```
So you could tell that the queue was released, but there weren't any commands to process.

File: src-built-in/adapters/indexedDBAdapter.js [hunk 2]
```diff
@@ -134,6 +137,15 @@ const IndexedDBAdapter = function () {
...
+		}	
+		while (this.queue.length) {
+			let action = this.queue.shift();
```
NinjaManatee: Can be `const`

src-built-in/adapters/indexedDBAdapter.js
```diff
@@ -178,6 +190,11 @@ const IndexedDBAdapter = function () {
 		Logger.system.debug("IndexedDBAdapter.delete for key=" + combinedKey);
 		console.debug("IndexedDBAdapter.delete for key=" + combinedKey);
 
+		if (!this.db) {
```
NinjaManatee: In general, I think I would put the queue pushes before the logging statements, so the delete of a key doesn't get logged twice, even though it only gets executed once.

src-built-in/adapters/indexedDBAdapter.js
NinjaManatee: I think clearCache is missing a queue push.

PR #171, opened 2018-09-26
Make a configurable "prompt on save" for workspace

amaroks:
**Resolves [#8526](https://chartiq.kanbanize.com/ctrl_board/18/cards/8526/details)**

- Add "Prompt me before saving" under Preferences > Workspaces.
- Get initial setting value from Storage instead of hardcoded value.
- Subscribe to "Prompt me before saving" preference changes.

**What to verify:**

- That the checkbox is there and persist changes on finsemble reloads
- That the prompt dialog appears if "Prompt me before saving"" is checked
- That the prompt dialog does not appear if "Prompt me before saving" is unchecked

**More details**
https://chartiq.kanbanize.com/ctrl_board/18/cards/8526/details


carterbs:
@amaroks can you resolve conflicts when you get back in. This seems useful.

PR #179, opened 2018-10-08
Added zoom popup a la Chrome

NinjaManatee:
Added a popup to show the current zoom level. Popup also has zoom in/reset/out buttons a la Chrome the user can use to zoom in and out.

Added a min/max zoom, and made all of the zoom settings configurable.


src-built-in/components/badBoy/index.html
```diff
@@ -1,35 +1,35 @@
...
-  <div class="bp3-button-group .modifier">
-    <a class="bp3-button bp3-icon-repeat" id="multi2" role="button">2 seconds</a>
-    <a class="bp3-button bp3-icon-repeat" id="multi5" role="button">5 seconds</a>
-    <a class="bp3-button bp3-icon-repeat" id="multi10" role="button">10 Seconds</a>
-  </div>
-  <br />
-
-  <pre id="message"></pre>
-</body>
+<!DOCTYPE html>
```
NinjaManatee: This file got into the PR somehow. Can you get it out again? I tried to `git checkout master src-built-in/components/badBoy` to fix it, but git said the file didn't exist.

configs/other/server-environment-startup.json
```diff
@@ -9,6 +9,6 @@
         "serverPort": 3375,
         "serverConfig": "http://localhost:3375/configs/openfin/manifest-local.json",
         "clientRoute": "http://localhost:3375/",
-        "channelAdapter": "e2o"
+        "channelAdapter": "openfin"
```
NinjaManatee: This change shouldn't be in this PR either.

src-built-in/preloads/zoom.js
```diff
@@ -2,38 +2,138 @@
 // This global will contain our current zoom level
 window.fsblZoomLevel = 1;
 
+// Global variable with the timeout for the zoom
+window.zoomTimeout = 1000;
```
NinjaManatee: I'd like to add config for this too, but I'm not sure where that should be.

configs/other/server-environment-startup.json
NinjaManatee: I think the issue is that you are merging with master rather than develop, but the PR is against develop, so you are picking up changes that shouldn't be there.

src-built-in/preloads/zoom.js
NinjaManatee: NVM, looks like I already did it (see zoomConfigHandler). I put it under Window Manager.

sgd2z:
@kriswest this doesnt seem to work with the plus on the numeric keypad. It works with the plus next to backspace

kriswest:
I had a go at adding an extra hotkey registration for 'Ctrl' + '+' to make the keypad + work:
Line 176-177 of zoom.js:
```	
	FSBL.Clients.HotkeyClient.addBrowserHotkey(["ctrl", "="], zoomIn);
	FSBL.Clients.HotkeyClient.addBrowserHotkey(["ctrl", "+"], zoomIn);
```
, but it doesn't work (it fires twice on 'Ctrl' + '=' instead). This is because  finsemble's /src/common/keyMaps.json maps a couple of number line keys down, i.e. '+' and '_' are mapped down to '=', and '-'.
Line 184 - 190 of finsemble /src/common/keyMaps.json
```
		"+": "=",
		"-": "-",
		"/": "/",
		";": ";",
		"=": "=",
		",": ",",
		"_": "-",
```

Underscore (_) is not a problem as there is no separate key for that, but the + key should not be mapped down - there's no other dodgy key mapping like that in the dictionary. This was probably a case of someone optimizing too early. 

For now I'd leave the zoom preload as is and instead fix the key mapping in finsemble and circle back to this after thats merged. @sgd2z if thats ok with you, approve this and I'll sub a quick PR for changing that mapping.

kriswest:
@NinjaManatee Added support for Ctrl + mousewheel zooming as this hasn't been merged yet and Natwest devs asked for it. Works really nicely.

PR #182, opened 2018-10-09
9562 New app launcher component

amaroks:
**Resolves [#9562](https://chartiq.kanbanize.com/ctrl_dashboard//cards/9562/details)**

-- This is in progress --


src-built-in/components/appLauncher2/src/components/AppDefinition.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+	render() {
+		const app = this.props.app
+		return (
+			<div className="app-item">
+				<div className="app-item-title">{app.friendlyName}</div>
+					<div className="app-item-tags">
+						{
+							app.tags.map((tag, index) => {
+								return <span key={index}>{tag}, </span>
```
carterbs: add a check here. `if(index !== app.tags.length)`, then render the `,`.

src-built-in/components/appLauncher2/src/components/Content.jsx
```diff
@@ -0,0 +1,39 @@
...
+		super(props)
+		this.state = {
+			filter: {
+				text: ''
+			}
+		}
+	}
+
+	filterApps() {
+		const apps = this.props.activeFolder.appDefinitions
```
carterbs: Once you get to making this filter work, you'll want your filterSort component to set a value on the store; this component will listen for changes on that field. When the value changes, this component renders with the newly filtered list.

src-built-in/components/appLauncher2/src/components/LeftNav.jsx
```diff
@@ -0,0 +1,48 @@
+import React from  'react'
+import FoldersList from  './FoldersList'
+
+const bottomEntries = [
+	'New folder',
+	'New dashboard',
+	'App catalogue'
```
carterbs: Go ahead and use american english here. "App Catalog"

src-built-in/components/appLauncher2/src/components/LeftNav.jsx
```diff
@@ -0,0 +1,48 @@
...
+				  			className={className} key={index}>
+				  				{folder.name}
+				  			</div>
+				  			)
+				  	})
+				  }
+			  </div>
+			  <div className="bottom">
+				  {
+				  	bottomEntries.map((entry, index) => {
```
carterbs: There are two schools of thought here, and I don't know where I come down. Maybe @mesner has an opinion.

You can do what you're doing here. Inline array iteration spitting out DOM elements. Alternatively, you can break these out into functions, and just do this: `{this.renderBottomEntries()}`.

I'm not sure which is better but wanted to make you aware of both approaches. I tend to do what you're doing here.

src-built-in/components/appLauncher2/src/components/AppDefinition.jsx
amaroks: Thanks for pointing to this, did it slightly different 

> {app.tags[index+1] ? `${tag}, ` : `${tag}`}

src-built-in/components/appLauncher2/src/components/LeftNav.jsx
amaroks: I didn't know about the second approach, but I really like it, I have moved the top links to a function and the bottom one to own component, the only reason I didn't break it into more components is that I don't know much about what the store looks like. so I will revisit structure at a later point.

src-built-in/components/appLauncher2/src/components/AppDefinition.jsx
```diff
@@ -13,13 +13,15 @@ export default class AppDefinition extends React.Component {
...
-							app.tags.map((tag, index) => {
-								return <span key={index}>{tag}, </span>
-							})
-						}
-					</div>
+				<div className="app-item-tags">
+				{
+					app.tags.map((tag, index) => {
+						return <span key={index}>
+						{app.tags[index+1] ? `${tag}, ` : `${tag}`}
```
carterbs: why +1 here??

src-built-in/components/appLauncher2/src/components/Content.jsx
```diff
@@ -1,37 +1,77 @@
...
+	}
+	
+	onSort(error, data) {
+		this.setState({
+			sortBy: data.value
+		})
+	}
+
+	componentWillMount() {
+		getStore().addListener({field: 'activeFolder'}, this.onActiveFolderChanged)
```
carterbs: Let's go ahead and use a variable here.

```
let store = getStore();
store.addListener...
```

src-built-in/components/appLauncher2/src/components/FilterSort.jsx
```diff
@@ -1,25 +1,42 @@
...
+			search: event.target.value
+		}, () => {
+			getStore().setValue({
+				field: 'filterText', 
+				value: this.state.search
+			})
+		})
+	}
+
+	onItemClick(item) {
```
carterbs: Make sure you install some kind of linter. the single quotes will get replaced with double quotes, formatting will fix itself. I highly recommend VSCode.

src-built-in/components/appLauncher2/src/components/Content.jsx
amaroks: Is there a good reason for creating unnecessary variables? Im a believer in "unnecessary variables are bad for your code"

src-built-in/components/appLauncher2/src/stores/StoreActions.js
```diff
@@ -0,0 +1,55 @@
+import {getStore} from './LauncherStore'
+
+export default {
+	getFolders,
```
carterbs: I like this

src-built-in/components/appLauncher2/src/components/AppDefinition.jsx
carterbs: You are correct. didn't even see the ternary.

src-built-in/components/appLauncher2/src/components/Content.jsx
carterbs: I'm a believer in the rule of 3. If I see an object being used more than 3 times, I'd like to see a variable. For me, variables really help when I'm reading the code. I see `store = getStore()`, and I can file that away mentally.

Then, when I see store later down in the code, I know it's the return value of getStore.

https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming) for a little context

PR #188, opened 2018-10-12
[9791] fix sizing of toolbar pins with icons

kriswest:
**Resolves #9791**
- Set display: inline-flex on .finsemble-toolbar-button

**What to verify:**
- Check that toolbar pins with icons are displayed correctly


NinjaManatee:
Looks like it works to me.

Before: 
Before fix applied
After:
After fix applied

PR #190, opened 2018-10-15
fix css issue with toolbar pins

sgd2z:
**Resolves #9791**

- Added display:inline-flex to fix css issue with toolbar pins that had both icons and text

**What to verify:**
- Welcome Component pin shows both icon and text.


sgd2z:
Duplicate of - https://github.com/ChartIQ/finsemble-seed/pull/188

PR #196, opened 2018-10-19
get proper bounds instead of bad bounds from componentState

sgd2z:
**Resolves #9908**

- Use getBounds instead of corrupted "window-bounds" in componentState to get bounds

**What to verify:**
Toolbar doesnt vanish on restart in Finsemble electron


sgd2z:
I have no issues with restoring as undocked toolbar. I deleted cache, did dev: fresh to make sure I'm not cached and my toolbar restores just fine. Do you have errors or anything?

PR #199, opened 2018-10-22
9876 double click toolbar

carterbs:
**Resolves 9876**

- On maximized, restore the window. This causes a flash, but doesn't allow the toolbar to take up all of your monitor's space.

**What to verify:**

- The toolbar doesn't maximize when double clicking the drag handle.

**Card**
https://chartiq.kanbanize.com/ctrl_board/18/cards/9876/details

**Finsemble PR**
https://github.com/ChartIQ/finsemble/pull/928

### DIFF
--- a/src-built-in/components/toolbar/stores/toolbarStore.js
+++ b/src-built-in/components/toolbar/stores/toolbarStore.js
@@ -171,7 +171,6 @@ class _ToolbarStore {
 			done();
 		});
 		let onBoundsSet = (bounds) => {
-			debugger
 			bounds = bounds.data ? bounds.data : bounds;
 			self.Store.setValue({ field: "window-bounds", value: bounds });
 			FSBL.Clients.WindowClient.setComponentState({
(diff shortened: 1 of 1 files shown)


src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -171,16 +171,19 @@ class _ToolbarStore {
...
 				value: bounds
 			}, Function.prototype);
 		}
-
-		FSBL.Clients.WindowClient.finsembleWindow.addListener("bounds-change-end", onBoundsSet)
+		let restoreWindow = (e) => {
+			finsembleWindow.restore();
+		}
+		//Immediately restore on maximize.
+		finsembleWindow.addListener("maximized", restoreWindow);
```
thorsent: @carterbs I think this is what's happening? "to prevent double click from maximizing, we instantly restore the window whenever it is maximized".

If so, there ought to be a comment stating explicitly what is happening here and why (eliminate campfire knowledge).

We might want to turn this into a window capability (allowMaximize:false, allowMinimize:false).

thorsent: I think you could just commit directly to the branch.

PR #204, opened 2018-10-25
10023 Fix linkerWindow positioning

carterbs:
**Resolves 10023**

- Offset by the linkerButton's parent instead of the linker button. This prevents the strange positioning mentioned in the card.

**What to verify:**

- Click linker button.
- Linker window shows in proper place.


sgd2z:
Creating new one with these specific commits cherry-picked

PR #206, opened 2018-10-26
10054 commenting out fetch to non-existent logout endpoint and fixing anti-…

kriswest:
…pattern (calling server-side logout before saving workspace)

**Resolves CS: #10053  PD: #10054**

It wasn't clear to a First Derivatives developer how to implement a log-out function (logs a user out and restarts without closing Finsemble and forcing user to relaunch). Our FileMenu component contains an example which sends a request to a server-side function, which doesn't actually exist and hence causes a 404, is not labelled as such and is potentially an anti-pattern anyway (it sends the logout request *before* attempting to save the workspace). 

Further the FSBL.restartApplication() and FSBL.shutdownApplication() functions are undocumented - a note was sent to Josh to correct this separately


d4hines:
This should be merged into `develop`.

d4hines:
I've resolved conflicts. It's ready to merge. I'll make a KB card.

PR #208, opened 2018-10-26
10081 fixing race in IndexedDBAdapter that occurs on first use with a clean…

kriswest:
… cache

**Resolves #CS: 10081**
https://chartiq.kanbanize.com/ctrl_board/21/cards/10081/details

- Introduces an initialized flag which controls request queuing (instead of checking for the DB connection) which is set after any schema changes are applied.

**What to verify:**
- Clear your OpenFin cache, 
- ensure seed is configured for the IndexedDBAdapter, 
- start finsemble.
- check central logger for related errors, there should be none.

### DIFF
--- a/src-built-in/adapters/indexedDBAdapter.js
+++ b/src-built-in/adapters/indexedDBAdapter.js
@@ -13,6 +13,11 @@ Logger.start();
  */
 const SCHEMA_VERSION = 1;
 
+/**
+ * Flag indicating that the IndexedDB connection is initialized. Will be set in releaseQueue after the connection has been created and any schema version changes applied.
+ */
+let initialized = false;
+
 // #region 
 /** 
  * PolyFill IDBKeyRange for a key prefix search. 
@@ -147,6 +152,7 @@ const IndexedDBAdapter = function () {
 	 * @private
 	 */
 	this.releaseQueue = () => {
+		initialized = true;
 		Logger.system.debug(`IndexedDBAdapter.releaseQueue: ${this.queue.length} commands`);
 		console.debug(`IndexedDBAdapter.releaseQueue: ${this.queue.length} commands`);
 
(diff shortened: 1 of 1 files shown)


src-built-in/adapters/indexedDBAdapter.js
```diff
@@ -147,6 +152,7 @@ const IndexedDBAdapter = function () {
 	 * @private
 	 */
 	this.releaseQueue = () => {
+		initialized = true;
```
d4hines: @kriswest, forgive me if this is a dumb question; a lot of this is still opaque to me. What's the connection between `this.releaseQueue` being called and the db's initialization? How do you know it's finished by by the time this is called?

d4hines:
We can't merge straight into master. We should reopen this PR for `develop`.

kriswest:
releaseQueue is called when the dB connection is ready, either after a
normal connection is successfully opened OR try connection is opened and
the schema is updated.

The error reported was that, with a clear cache (and hence no schema in the
IndexedDB) that some early requests to the storage adapter were failing
(with a message indicating a schema update was underway). This was due to
us not waiting for that to happen (the connection was available, but the
schema update was still underway). This PR alters things such that requests
are queued until the queue is released (which happened as soon as the
startup is truly finished). Resolves the issue for me at any rate.

I can't talk you through the adapter code next week if that would help.

K

d4hines:
Ok, sorry about this then - I got what seemed to be conflicting
information, but I'll have Terry/Ian clarify.

kriswest:
No problem Daniel - I think we're still working this out and will be
meeting about it next week. I've been on the road for a week and half, so
if anyones got the wrong info its me ;-)

K



Kris West
Senior Solutions Architect
p: +1 800 821 8147  m: +44 (0)7715104292
w: www.chartiq.com  e: kris@chartiq.com

src-built-in/adapters/indexedDBAdapter.js
sgd2z: Looks like the difference is if it follows the "onupgradeneeded" path. in this case it only gets initialized after the oncomplete as opposed to the previous case where this.db was getting set before that.

thorsent: Duh, race condition. Good catch.

PR #210, opened 2018-10-30
8627 Example config prop for showing dev console on component launch

clausjs:
**Resolves [8627](https://chartiq.kanbanize.com/ctrl_board/18/cards/8627/details)**

- Adds a config prop which will open dev console on component startup[

**What to verify:**

- If the 'showDevConsoleOnVisible' prop is true for a component's config, expect the dev console to launch alongside the component


d4hines:
Having confirmed it's working, we'll close this.

PR #214, opened 2018-10-30
9277 tab styles

carterbs:
**Resolves #9277**

- Made CSS match what was in Nicolai's mockup. Reworked the logo component.

**What to verify:**

- Needs to look good.
- Shouldn't break the ability to move a window.

### DIFF
--- a/assets/css/_theme.css
+++ b/assets/css/_theme.css
@@ -50,7 +50,7 @@
 
     /* Window Title bar */
     --titleBar-height: 25px;
-    --titleBar-font-size: 14px;
+    --titleBar-font-size: 11px;
     --titleBar-icon-font-size: 10px;
     --titleBar-active-font-color: var(--primary-font-color);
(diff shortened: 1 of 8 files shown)


DanNicolai:
This is looking good and functioning well. A few requests:

-  Let's get rid of the floating border when there's only one tab. It's awkward looking and floats awkwardly when resizing the window. 

- The drop area for tabs has shrunk, and now you can only insert tabs over other tabs. If I recall correctly, the old behavior allowed for dragging tabs anywhere on the header (and displayed a full-header scrim). If possible I'd like to have this restored.

- When there's only one tab, I would like it for the tab to appear as 125px when hovered over. Currently, it doesn't appear fixed at that width until you drag a tab into it and/or create a multiple tabbed window.

thorsent:
This never went through design and we want to do an overhaul of tabbing. It's been decided to kill this until we have the ability to revisit in depth.

PR #216, opened 2018-10-31
10271 no value should be undefined, rather than an empty object

NinjaManatee:
**Resolves #10271**

- return undefined instead of {} when there isn't a value in the IndexedDB

**What to verify:**

Before https://github.com/ChartIQ/finsemble/pull/961

- Enable indexedDB adapter
- Create new workspace
- Try to switch workspaces (fails before this fix and the above PR)

### DIFF
--- a/src-built-in/adapters/indexedDBAdapter.js
+++ b/src-built-in/adapters/indexedDBAdapter.js
@@ -288,7 +288,8 @@ const IndexedDBAdapter = function () {
 		const request = objectStore.get(combinedKey);
 
 		request.onsuccess = (event) => {
-			const data = event.target.result && event.target.result.value ? event.target.result.value : undefined;
+			let data;
+			if(event.target.result) data = event.target.result.value;
 
 			Logger.system.debug("IndexedDBAdapter.get for key=" + combinedKey + " data=", data);
 			console.debug("IndexedDBAdapter.get for key=" + combinedKey + " data=", data);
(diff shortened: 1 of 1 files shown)


src-built-in/adapters/indexedDBAdapter.js
```diff
@@ -288,7 +288,7 @@ const IndexedDBAdapter = function () {
 		const request = objectStore.get(combinedKey);
 
 		request.onsuccess = (event) => {
-			const data = event.target.result && event.target.result.value ? event.target.result.value : {};
+			const data = event.target.result && event.target.result.value ? event.target.result.value : undefined;
```
thorsent: @NinjaManatee it feels like this code should be:
```
let data;
if(event.target.result) data = event.target.result.value
```

Currently you're forcing "null" to be "undefined". I think this is an artifact of how the code has evolved from originally forcing it to {}.

NinjaManatee: Yeah, I did the minimum change, but didn't consider converting null to undefined. I think this would do what you want as well:
```
const data = event.target.result ? event.target.result.value : undefined;
```

If value exists, null or otherwise, its value would be used, but if result doesn't exist the default value is undefined. The assumption is that target will always exist, but I'm not 100% sure whether that is a valid assumption.

mhmcclung:
Side Note: the errors in the storage service are extraneous (a side effect of FSBL getting pulled into a service and on our to-do list to resolve).  Also, the default application/config.js in this branch still uses local storage.  I only temporarily changed to IndexedDB to verify Finemble was working fine.

PR #217, opened 2018-10-31
Use persistent store in App launcher

amaroks:
**Resolves #10170**

- Update store actions to deal with store data asynchronously
- Switch from calling store actions synchronously to ES7 async/await
- Wrap StoreModel.getValue in a function that returns a promise
- Add default and foundation data to configs/applications/config.js

**What to verify:**

- That you can still use app launcher the same as before
- That when you add a new folder or add tags they persist in next restart/reload
- That changes persist when you reorder folders or drag and drop apps
- That you don't see any react/js errors in console


configs/application/config.json
```diff
@@ -13,6 +13,95 @@
 		}
 	},
 	"servicesConfig": {
+		"distributedStore": {
+			"initialStores": [
+				{
+					"name": "Finsemble-AppLauncher-2",
```
rcsharp5: Wrong config

src-built-in/components/appLauncher2/src/components/AppActionsMenu.jsx
```diff
@@ -30,12 +31,20 @@ export default class AppActionsMenu extends React.Component {
 	}
 
 	//TODO: Implement handlers
-	onAddToFavorite() {
-		const favorite = storeActions.getFolders().find((folder) => {
-			return folder.name === 'Favorites'
-		})
+	async onAddToFavorite() {
+		const favorite = await storeActions.getFavoriteFolder()
```
rcsharp5: Why is this async?

src-built-in/components/appLauncher2/src/components/Content.jsx
```diff
@@ -50,15 +64,15 @@ export default class Content extends React.Component {
 		})
 	}
 
-	onActiveFolderChanged(error, data) {
+	async onActiveFolderChanged(error, data) {
```
rcsharp5: This is going to be a general comment for this file. You should not need all of the asyncs. If you're using the store state should be updated from a listener.

rcsharp5: You already have a listern for this section `store.addListener({ field: 'activeFolder' }, this.onActiveFolderChanged)`. No need for async.

src-built-in/components/appLauncher2/src/components/FolderActionsMenu.jsx
```diff
@@ -0,0 +1,76 @@
...
+	}
+	toggleMenu() {
+		this.setState({
+			isVisible: !this.state.isVisible
+		});
+	}
+	renameFolder() {
+		this.props.renameFolder(this.props.folder.name);
+	}
+	async launchAll() {
```
rcsharp5: Don't need async here

src-built-in/components/appLauncher2/src/stores/StoreActions.js
```diff
@@ -1,9 +1,12 @@
 import { getStore } from './LauncherStore'
 
 export default {
```
rcsharp5: Another general comment. Lets use listeners for all of this data.

src-built-in/components/appLauncher2/src/components/AppActionsMenu.jsx
amaroks: Because storeActions.getFavoriteFolder() returns a promise and we don't want then(function ...)

src-built-in/components/appLauncher2/src/components/Content.jsx
amaroks: We can't, activeFolder is a string, while content requires the active folder object. The eventListener will return the name of the folder only.

src-built-in/components/appLauncher2/src/components/FolderActionsMenu.jsx
amaroks: @rcsharp5 I'm not sure why you think there is no need for it there. Well I see @clausjs  is using await storeActions.getFolders() so definitely he needs async there. unless he passes and gets the folders from props or get them in componentWill/DidMount.

PR #219, opened 2018-10-31
Fixes to address most bugs in 10202

clausjs:
- Addresses fixes/bugs here https://chartiq.kanbanize.com/ctrl_board/18/cards/10202/subtasks


clausjs:
The styling of the folder list context menu is off. Nicolai will fix tomorrow

PR #224, opened 2018-11-07
Revert "10081 fixing race in IndexedDBAdapter that occurs on first use with a clean…"

thorsent:
Reverts ChartIQ/finsemble-seed#208


thorsent:
Will merge through release cycle.

PR #225, opened 2018-11-08
[App Launcher] Text and Icons

DanNicolai:
**Resolve issue [10383](https://chartiq.kanbanize.com/ctrl_board/18/cards/10383/details)**

**Description of change**
- Restored text truncation in the appLauncher
- Changed pin icons to ff-favorite (stars) to reflect consistency with new appLauncher. Adjusted some hover-states for further consistency.

**Description of testing**
- Put in a large amount of text anywhere in the appLauncher. Text should not inappropriately overflow or stretch the interface.
- Check workspace menu for star icons instead of pins. Make sure they possess hover states.

### DIFF
--- a/assets/css/_menus.css
+++ b/assets/css/_menus.css
@@ -104,29 +104,43 @@ body.menu {
 }
 
 .ff-delete:hover {
-    color: var(--menu-font-color);
-    background-color: var(--menu-delete-background-hover-color);
+    color: var(--menu-delete-background-hover-color);
+    /* background-color: var(--menu-delete-background-hover-color); */
     border-radius: 4px;
 }
 
 .ff-pin:hover {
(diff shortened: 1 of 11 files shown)


amaroks:
@DanNicolai I don't see star icons instead of pins anywhere, am I missing something? Also I don't think we should make any changes outside App Launcher and App Catalog and then merge to our epic.

DanNicolai:
@amaroks the Star icons are in the Workspaces menu on this branch. This was intentional so that it would match the star icons in the appLauncher. I wanted to squash this inconsistency while there was still a chance, and I did talk it over with @mesner before putting these changes in. 

If we still don't want to  do this, that's fine, but I feel like it's only going to be harder to make this change in the future.

PR #227, opened 2018-11-08
[AppCatalog] 10202 final mvp bug fixes

clausjs:
**Resolves issue [10202](https://chartiq.kanbanize.com/ctrl_board/18/cards/10202/subtasks)**

**Description of change**
-Fixes bugs mentioned in 10202 subtasks.
-Specifically addresses open/remove buttons, carousel operability, and tag search to be AND instead of OR

**Description of testing**
- Addresses all checked subtasks from 10202

### DIFF
--- a/assets/css/_theme.css
+++ b/assets/css/_theme.css
@@ -250,6 +250,7 @@
     --tag-font-color: var(--primary-font-color);
     --tag-font-weight: var(--primary-font-weight);
     --tag-close-icon-hover-color: var(--secondary-negative-color);
+    --tag-checkmark-icon-color: var(--green);
 
 
     /* App Menu */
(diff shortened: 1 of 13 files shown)


DanNicolai:
- Sort by favorites still doesn't appear to work
- Can't quite test if sorting by Recent is working because of lack of real apps. Remove the option if you don't feel this will work.
- Also can't quite tell if "View Info" is working because of lack of real apps. Going to wait to see if this  comes true.
- Reordering folders bug still prevalent, but fix is sitting in 10383. Merge that in?

amaroks:
@DanNicolai First and last were all fixed in other PRs.

clausjs:
@amaroks what about the others?

PR #228, opened 2018-11-08
Fixes tags menu visually indicate what tags are active

clausjs:
**Resolves issue [10445](https://chartiq.kanbanize.com/ctrl_board/18/cards/10445/details)**

**Description of change**
- Added font icon to tags menu for app catalog and new app launcher that will indicate when a tag is active or not
- Instead of the tags menu simply adding tags, it can now toggle tags. If its already applied and is clicked it will be removed and vice versa

**Description of testing**
- Verify that you can add a tag in app catalog and app launcher 2. When the tag is added, a check mark is added next to the tag that means its enabled. Clicking any tags with check marks next to them will also remove that tag.


DanNicolai:
Modified and tested. This should be good to go.

PR #230, opened 2018-11-09
[App Catalog] Fix Carousels with <> 4 cards. Handle  app results when cards on a pa…

clausjs:
**Resolves issue [10447](https://chartiq.kanbanize.com/ctrl_board/18/cards/10447/details)**

**Description of change**
- Adjusts carousels to allow for (and console warn) a user that a carousel should have at least 4 cards for optimal displaying
- Adjusts AppResults page when filtering by tags to all for rows of less than 4 for when filtered results are not divisible by 4.

**Description of testing**
- (Requires a specific branch of fdc-appd) Upon opening the app one carousel should seem off as it only has 2 cards, the other will have 5 cards and allow for previewing the disabled left and right arrows when the end of a list is reached.

### DIFF
--- a/src-built-in/components/appCatalog2/src/components/AppResults.jsx
+++ b/src-built-in/components/appCatalog2/src/components/AppResults.jsx
@@ -55,11 +55,6 @@ const AppResults = props => {
 			let card = cardsForShowcase[i];
 			let name = card.title || card.name;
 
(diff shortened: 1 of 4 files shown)


DanNicolai:
Working well, just need to make sure that the tag fix also eventually applies to Catalog (which you informed me is on another branch)

PR #232, opened 2018-11-09
[AppLauncher] Launch apps with single click, tag filter using AND instead or OR

clausjs:
**Resolves issues**
[10405](https://chartiq.kanbanize.com/ctrl_board/18/cards/10383/subtasks)
[10436](https://chartiq.kanbanize.com/ctrl_board/18/cards/10383/subtasks)

**Description of change**
- Apps now launch from a single click on the appLauncher (will not launch apps when the context menu or tags are clicked)
- Tag search is now AND not OR. Only apps containing _all_ currently filtered tags will show

**Description of testing**
- Confirm that multiple active tags only shows apps that contain all those tags, and that a single click will launch an app

### DIFF
--- a/src-built-in/components/appLauncher2/src/components/AppActionsMenu.jsx
+++ b/src-built-in/components/appLauncher2/src/components/AppActionsMenu.jsx
@@ -33,7 +33,11 @@ export default class AppActionsMenu extends React.Component {
 		document.removeEventListener('mousedown', this.handleClickOutside);
 	}
 
-	toggleMenu() {
+	toggleMenu(e) {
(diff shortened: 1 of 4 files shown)


DanNicolai:
Working well, just need to make sure that the tag fix also eventually applies to Catalog (which you informed me is on another branch)

DanNicolai:
Bug: the menu doesn't close when you launch the Getting Started component

Otherwise, seems to be working well.

amaroks:
Looks good, will merge this and we will take care of the menu auto hide in another PR.

PR #176, opened 2018-10-03
Preload window title bar

amaroks:
Solves ticket [#8922](https://chartiq.kanbanize.com/ctrl_board/18/cards/8922/comments)

PR #236, opened 2018-11-12
8922 preload titlebar

sgd2z:
**Resolves issue [8922](https://chartiq.kanbanize.com/ctrl_board/18/cards/8922/details)**

**Description of change**
- Redid Mody's changes to the new seed develop. See https://github.com/ChartIQ/finsemble-seed/pull/176
- Also requires Finsemble PR - https://github.com/ChartIQ/finsemble/pull/980

**Description of testing**
- Verified that the titlebar preloads and functions on the titlebar work.
- Note that existing components when reloaded after this change will not have titlebars. So need to clear cache or close those.


src-built-in/components/windowTitleBar/src/windowTitleBar.jsx
```diff
@@ -1,434 +1,11 @@
...
-	storeExports.initialize(function () {
-		HeaderActions = storeExports.Actions;
-		windowTitleBarStore = storeExports.getStore();
-		ReactDOM.render(<WindowTitleBar />, document.getElementById("FSBLHeader"));
-	});
-}
+function init() {
+	if (window.headerLoaded) return;
+	window.headerLoaded = true;
+	require("./windowTitleBar2.jsx");
```
carterbs: rename this file to `entry.jsx` and keep the other as windowTitleBar.jsx. Then change the webpack. titlebar and titlebar2 is   gross.

carterbs:
@sgd2z 

> Note that existing components when reloaded after this change will not have titlebars. So need to clear cache or close those.

Is this an absolute? We can't live with that. Citadel won't be happy with this. Why can't we add the preload in there if it doesn't exist?

carterbs:
@sgd2z followup question, why is there an entry file at all?

carterbs:
Was done more recently because of electron support. closing.

PR #237, opened 2018-11-12
10551 - Clean localStorageAdapter.js

d4hines:
**Resolves issue [10551](https://chartiq.kanbanize.com/ctrl_board/18/cards/10551/details)**

**Description of change**
- Removed unsanitized call to RegExp, replacing it with `string.startsWith()`.
  - For a performance analysis, see https://jsperf.com/string-startswith-vs-regex-caret
  - Gist is that `string.startsWith()` can be slower, but only by a factor of at most 2.

**Description of testing**
- I verified the `this.keys()` method still returned the appropriate results after the fix.


d4hines:
I'm not really sure if this is what you wanted, but I'll give it a shot. I added some validation and docs to `this.keys`. How does that look? Or did you mean refactor the whole module?

carterbs:
Looks great. Ready for terry

PR #240, opened 2018-11-13
10383 favorite toolbar pins and folder renaming

clausjs:
**Resolves issue [10406](https://chartiq.kanbanize.com/ctrl_board/18/cards/10383/subtasks)**

**Description of change**
- Removes the folder rename dialog when a repeated name is encountered. Instead the submission goes through and automatically appends "(1)" to the name
- When an app is added/removed from favorites a pin is added/removed from the toolbar

**Description of testing**
- Ensure that adding/removing an app from favorites in the launcher adds/removes a pin from the toolbar.
-  Ensure that when trying to rename a folder, a repeated name cannot be used


DanNicolai:
Tested and good to go

PR #243, opened 2018-11-13
[AppLauncher] [AppCatalog] View app info

amaroks:
**Resolves issue [#10407](https://chartiq.kanbanize.com/ctrl_board/18/cards/10407/details)**

**Description of change**
- Modified Welcome component id to match one in appd
- Transmit `viewApp` whenever user clicks on `view info`
- Listen for `viewApp` events in appCatalog app.jsx

**Description of testing**
Open launcher and catalog, open welcome app actions menu and click on `view info`, app catalog should be brought to front and switched to welcome app info page.

### DIFF
--- a/configs/application/config.json
+++ b/configs/application/config.json
@@ -28,7 +28,7 @@
 									"apps": {
 										"1": {
 											"name": "Welcome",
-											"appID": "1"
+											"appID": "welcome-comp"
 										},
 										"2": {
 											"name": "Getting Started Tutorial",
@@ -53,8 +53,8 @@
 							}
 						},
 						"appDefinitions": {
-							"1": {
-								"appID": "1",
+							"welcome-comp": {
+								"appID": "welcome-comp",
 								"name": "Welcome",
 								"source": "FDC3",
 								"tags": [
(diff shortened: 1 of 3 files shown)


DanNicolai:
@amaroks this only seems to work if the catalog is already open. Otherwise, the catalog will open but stays on its homepage with the carousels.

Can you make it so that if the catalog isn't open, it still redirects to the info page?

clausjs:
@DanNicolai test again but _only_ using the Welcome Component. The catalog is a list of prospective apps (dummy data for now) where as the launcher is a list of real added, launchable apps. The Welcome Component is really the only thing they share

DanNicolai:
Tested again using only welcome component but got the same result:
1) Open Launcher
2) Open Welcome component's context menu, select "View Info"
3) Catalog opens, but stays on the carousel page. It should be on the Showcase page.

DanNicolai:
Retested and is working properly now.

clausjs:
@amaroks I added a delay from the callback of spawning an app catalog and calling the router to transmit a message. This gives enough time to initialize the component but is small enough that when the catalog is already spawned, the delay will not feel too long. There might be a better way to do this, but you should have a look and see if you approve

PR #245, opened 2018-11-14
10628 uiux5030 spinner

d4hines:
**Resolves issue [10628](https://chartiq.kanbanize.com/ctrl_board/18/cards/10628/details)**

**Description of change**
Added a spinner in front of the item in the workspace management menu corresponding to the workspace currently being loaded.

**Description of testing**
-How did you verify that your code satisfies the card's requirements?
I switched between two large workspaces back and forth 5-6 times 

**What to Review**
@BradleyCarter, I think you added a callback that conflicted with my stuff, but I don't think it was necessary. Please look over and make sure I've implemented the behavior you want in `workspaceManagementMenuStore.js`.


src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
```diff
@@ -326,19 +332,21 @@ Actions = {
 	 * Asks the user if they'd like to save their data, then loads the requested workspace.
 	 */
 	switchToWorkspace: function (data) {
-		if (switching) return;
-		switching = true;
+		if (Actions.getIsSwitchingWorkspaces()) return;
+		Actions.setIsSwitchingWorkspaces(true);
```
thorsent: @d4hines you're changing from a simple boolean to using a store. That fundamentally complicates the logic, opening up the possibility of introducing race conditions. This task didn't require changing this code.

src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
```diff
@@ -375,13 +383,11 @@ Actions = {
 			}
 
 			//Switch is the last thing we do.
-			tasks.push(switchIt);
+			tasks.push(switchWorkspace);
 
 			async.waterfall(tasks, Actions.onAsyncComplete);
 		} else {
-			switchIt(() => {
-				switching = false;
```
thorsent: @d4hines Right here. change your state to hide the loader.

d4hines:
@thorsent, @BradleyCarter, I've switched back to the local var as requested. I've tested it out, and it works fine.

d4hines:
@BradleyCarter, we decided that using the local store is the best bet for this task. I left a comment about the downsides to excluding the local store and the consequence of the spinner not showing up if a workspace is selected from the toolbar pin instead of the menu.

This was my first time using `git revert`. I tried to carefully go through and make sure I didn't screw anything, but ,let me know if I missed anything.

PR #247, opened 2018-11-15
Changed server address to localhost.chartiq.com for websocket connection

NinjaManatee:
IAC should use localhost.chartiq.com for WSS connection rather than 127.0.0.1.


carterbs:
@NinjaManatee is this PR still necessary?

PR #249, opened 2018-11-15
[AppLauncher] New form for adding apps

amaroks:
**Resolves issue [#10714](https://chartiq.kanbanize.com/ctrl_board/18/cards/10714/details)**

**Description of change**
- Created a new form component
- Added a new store action for adding apps
- Used stores to save the visibility status of the form
- Update rendered list once app is added

**Description of testing**
- Reset finsemble
- Click on "New app" and make sure a form shows up
- Type a name, URL and click on submit button

Your new app should be visible in the list and the form should hides.


DanNicolai:
@amaroks I'm gonna see if I can resolve these styling issues with the dialog class.

Things that still need to be fixed on your end: 
- include the "https://" in the url submission so that users only have to type "www.google.com" to have the form validate. It's confusing right now as to why the form doesn't submit.
- Lowercase app names don't correctly sort in alphabetical order (sorry, wasn't able to observe this until apps were able to be created). 
- Newly added apps don't launch (not sure if this is a bug or just because of current config set up). 
- Newly added apps don't get a star when added to favorites
- Newly added apps don't get pinned to the toolbar when Favorited
- Newly added apps shouldn't have "View Info" in their context menu (because they're not in the catalog).
- Newly added apps need an option to "Delete App" in their context menu (because they're not in the catalog)

amaroks:
However, few points

- Apps with URL won't launch with current Finsemble version, needs an upgrade
- Why should we prepend https while the app URL could be a different unknown protocol?

DanNicolai:
@amaroks The search bar is still weird looking. Did you say that you and Brad had worked out a solution for that?

PR #253, opened 2018-11-19
10753 remove wsmm noise

d4hines:
**Resolves issue [10753](https://chartiq.kanbanize.com/ctrl_board/18/cards/10753/details)**

**Description of change**
Removed Logger noise by straightening out use of magic strings in error messages, moving them into constants and fixing mismatches.

**Description of testing**
- I tested to ensure that you could cancel from any of the following without any errors:
  - New Workspace
  - New Workspace with Overwrite
  - Save As
  - Save As with Overwrite

### DIFF
--- a/src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
+++ b/src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
@@ -527,7 +527,7 @@ Actions = {
 			callback();
 		} else {
 			//choice === cancel
-			callback(SAVE_DIALOG_CANCEL_ERROR);
+			callback(new Error(SAVE_DIALOG_CANCEL_ERROR));
 		}
 	},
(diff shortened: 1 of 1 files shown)


src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
```diff
@@ -565,7 +567,7 @@ Actions = {
 			if (response.choice === "affirmative") {
 				callback(null, workspaceName);
 			} else {
-				callback("Negative");
+				callback(NEGATIVE);
```
sgd2z: we should pick one of return (new Error(NEGATIVE)) or return(NEGATIVE). This is inconsistent

d4hines: Good catch! Done.

PR #255, opened 2018-11-20
[AppLauncher] 10754 ability to pin adhoc components

clausjs:
**Resolves issue [10754](https://chartiq.kanbanize.com/ctrl_board/18/cards/10754/details)**

**Description of change**
- When an adhoc component is added, FSBL is called to add the component

**Description of testing**
- Verify that an adhoc component added at runtime can pin and unpin from the taskbar using the 'favorite' feature

### DIFF
--- a/src-built-in/components/appLauncher2/src/stores/StoreActions.js
+++ b/src-built-in/components/appLauncher2/src/stores/StoreActions.js
@@ -195,6 +195,8 @@ function addApp(app = {}, cb) {
 	FSBL.Clients.LauncherClient.addUserDefinedComponent(data.apps[appID], (compAddErr) => {
 		if (compAddErr) {
 			//TODO: We need to handle the error here. If the component failed to add, we should probably fall back and not add to launcher
+			console.warn("Failed to add new app");
+			return;
 		}
 		// Save appDefinitions and then folders
 		_setValue('appDefinitions', data.apps, () => {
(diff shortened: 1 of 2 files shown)


DanNicolai:
Tested and is working. Assimilation version kept me from building but once it was change back to 2.0.3, everything worked well.

src-built-in/components/appLauncher2/src/stores/StoreActions.js
```diff
@@ -191,11 +192,16 @@ function addApp(app = {}, cb) {
 	}
 	data.folders[MY_APPS].apps[appID] = data.apps[appID]
 	data.folders[folder].apps[appID] = data.apps[appID]
-	// Save appDefinitions and then folders
-	_setValue('appDefinitions', data.apps, () => {
-		_setFolders()
-		cb && cb()
-	})
+	FSBL.Clients.LauncherClient.addUserDefinedComponent(data.apps[appID], (compAddErr) => {
+		if (compAddErr) {
```
clausjs: @amaroks We probably want to do something here. Maybe fall back and cancel the creation of the app entirely? If the component can't be added to FSBL, we shouldn't add it to the launcher

DanNicolai:
Can confirm that the app is still pinned after deleting. Once that's done this should be good to go.

PR #258, opened 2018-11-20
10326 Fix drag handler when window scrolls before titlebar load

clausjs:
**Resolves issue [10693](https://chartiq.kanbanize.com/ctrl_board/18/cards/10693/details)**

**Description of change**
- Changes the drag handler style changes to use 'setProperty' instead of accessing directly. This allows for the setting of property priority.
- Fixes a css type for window titlebars

**Description of testing**
- Create an adhoc component and spawn. Relead component and scroll before the page loads. Once the windowTitlebar is injected, make sure the window is still movable.

**NOTE: The old code was removing 5 pixels from the height to determine the negative margin of the drag handler, this caused a small deadzone near the tab/title. The 5 pixels was removed**

### DIFF
--- a/assets/css/_windowTitleBar.css
+++ b/assets/css/_windowTitleBar.css
@@ -7,7 +7,7 @@
     /* use below with window scrollbar hack */
     position: fixed;
     top: 5px!important;
-    margin-top: 0px!important;
+    margin-top: 0px !important;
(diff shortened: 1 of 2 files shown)


d4hines:
Still has intermittent titlebar issues, and seems to interfere with the welcome component's titlebar.

PR #259, opened 2018-11-20
10326 scroll before ready

clausjs:
**Resolves issue [10326](https://chartiq.kanbanize.com/ctrl_board/18/cards/10326/details)**

Adds a -25px top margin to a component that has a larger scrollHeight than window height. This is unfortunately only needed when the component is scrolled before the title bar is injected. Since the drag handler already has an important 0px top margin, we have to overwrite it with an important property _when its needed_. I have yet to solve _HOW_ to determine  if its needed.

### DIFF
--- a/assets/css/_windowTitleBar.css
+++ b/assets/css/_windowTitleBar.css
@@ -7,7 +7,7 @@
     /* use below with window scrollbar hack */
     position: fixed;
     top: 5px!important;
-    margin-top: 0px!important;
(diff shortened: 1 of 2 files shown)


carterbs:
@clausjs Need to know whether this PR is necessary

clausjs:
@BradleyCarter I assume not, the ticket I linked it to is in the backlog anyway, and doesn't even seem to have anything to do with this. Closing

PR #260, opened 2018-11-22
Gulp refactor

erik-ciq:
**There were three main goals in this PR.**

- [x] Replace callbacks with async / await 
- [x] Modularize the code
- [x] Increase readability, maintainability, and remove redundancy

**New modules were created:**

- [x] `gulpfile.js/buildSass` - Still a stub just like before
- [x] `gulpfile.js/envOrArg` 
- [x] `gulpfile.js/getFinsembleLinks` - returns objects with { path, name, version } for finsemble dependencies
- [x] `gulpfile.js/index.js` `gulpfile.js` -> `gulpfile.js/index.js` per gulp docs
- [x] `gulpfile.js/launchApplication` handles launching openfin or e20
- [x] `gulpfile.js/listableExecute` consumes an array of commands and runs them
- [x] `gulpfile.js/logToTerminal` one of the most heavily used functions I saw
- [x] `gulpfile.jssafeRequire` runs stat before requiring a module, abstracted to keep things DRY
- [x] `gulpfile.js/startServer` handles running the server
- [x] `gulpfile.js/tasks` composable tasks for gulp, more depth on this below
- [x] `gulpfile.js/verifyLinks.js` checks provided links, throws errors if they are missing
- [x] `./buildAngular.js` handles any angular work
- [x] `./buildWebpack.js` handles any webpack work

**Tasks:**
Tried to unify the atomic tasks that were being used. Pulled them into their own functions for re-usability and composition.

- [x] `pre`
- [x] `post`
- [x] `clean`
- [x] `build`  requires an `environment` arg
- [x] `serve` requires an `environment` arg
- [x] `launch` requires a `channelAdapter` and `manifest`
- [x] `buildServe` combines `build` and `serve`, requires an `environment` arg
- [x] `buildServeLaunch` combines `build`, `serve`, and `launch`, requires `environment` arg
- [x] `rebuildServeLaunch` combines `clean`, `build`, `serve`, and `launch`, requires `environment` arg
- [x] `rebuild` combines `clean` and `build`, requires `environment` arg
- [x] `serveLaunch` combines `serve` and `launch`, requires `environment` arg

**Description of testing**

- [x] Ensure running gulp tasks still works
- [ ] Test with extensions
- [ ] Test with angular configuration

I'd really like to update our gulpfile-extensions file as well if possible. Also curious to know if I can update any usages of this in the wild, my understanding is one of our customers uses this file. If I can update usages in the wild, I can unify things even more.

Would love to convert this to `.ts` and use interfaces to ensure we have tasks being called correctly. Maybe in a subsequent PR.


d4hines:
First of all, this is some great work! Everyone knows this stuff is a mess, but nobody's wanted to touch it so far. Thanks for sorting through all this!

I've gone through and cursorily reviewed the code, and it looks good. I've found a few bugs though. Some of them may have always been there, but at least a few are new, and in any case, since you've put this much work in, it should all work!

Where appropriate I left comments on the files. All other comments are here:

- The `tasks.js` file was missing `require('path')`. I added it - easy day.
- The `clean` task wasn't returning a promise, so it wasn't completing properly. I added `Promise.all` and it took care of it.
- The `clean` task isn't wiping the `./dist` folder. It's expecting that path to be passed in by the caller, but it never is.
- `prod:nolaunch` never exits. I'm not sure if it's supposed to be a "watch" thing, but when I try changing a file, I don't see the rebuild stuff either.
- `server` and `server:prod` fail for the reason in the comment above.

All of this should be pretty easily fixable. However, correct me if I'm wrong, but the main things theses scripts accomplish is: 
- Run webpack with some various configs based env.
- Optionally clean build artifacts at the right time.

Those two things are fairly simply accomplished with webpack configuration and npm scripts. So it's worth asking the question: do we need gulp scripts, or would it be simpler to just hammer out the kinks in webpack and go back to simple cli commands? If we want a lot of power and expressiveness in the build process, then by all means, gulp is the way to go for now. But if we can do without it, it's probably best that we do.

@NinjaManatee and @mesner should probably weigh in on these philosophical points.

erik-ciq:
@NinjaManatee  @mesner @d4hines this PR was incorrectly based off master, rather than develop. Closing this PR in favor of https://github.com/ChartIQ/finsemble-seed/pull/261

PR #270, opened 2018-11-30
10905 e2o fix

erik-ciq:
**Resolves issue [10905](https://chartiq.kanbanize.com/ctrl_board/18/cards/10905/details)**

**Description of change**
- [x] Update path of electron from node_modules to be correct.

**Description of testing**
Electron launching still works properly, but without requiring an additional npm install within `./node_modules/@chartiq/e2o`

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -382,7 +382,7 @@
 			});
 
 			let e2oLocation = "node_modules/@chartiq/e2o";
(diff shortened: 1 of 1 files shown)


carterbs:
@erik-ciq let me know if this needs to exist anymore. I think we're okay now.

PR #273, opened 2018-12-05
11024 allow electron debugging

sgd2z:
**Resolves issue [11024](https://chartiq.kanbanize.com/ctrl_board/18/cards/11024/details)**

**Description of change**
- new channelAdapter for debugging

**Description of testing**
- I Can debug the main process!!

### DIFF
--- a/configs/application/config.json
+++ b/configs/application/config.json
@@ -99,12 +99,6 @@
 					"storageService",
 					"searchService"
 				]
-			},
-			{
-				"agentLabel": "dockingService",
-				"services": [
-					"dockingService"
-				]
 			}
 		]
 	},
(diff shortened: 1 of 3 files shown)


configs/application/config.json
```diff
@@ -102,12 +102,6 @@
 					"storageService",
 					"searchService"
 				]
-			},
```
NinjaManatee: Is this change needed for this PR or is it something that has crept in?

configs/other/server-environment-startup.json
```diff
@@ -3,7 +3,9 @@
         "serverPort": 3375,
         "serverConfig": "http://localhost:3375/configs/openfin/manifest-local.json",
         "clientRoute": "http://localhost:3375/",
-        "channelAdapter": "openfin"
+        "channelAdapter": "openfin",
+        "e2odebug": true,
```
NinjaManatee: I wonder if it might be worthwhile to simplify the config. Maybe just have one property `"e2odebug"` where if it is true, it doesn't break on start, but if it is set to "break", then it will. 
```
        "channelAdapter": "openfin",
        "e2odebug": true,
```
Or
```
        "channelAdapter": "openfin",
        "e2odebug": "break",
```
Thoughts?

PR #275, opened 2018-12-06
11022 FOR REVIEW ONLY: Logging export service

kriswest:
**Resolves issue [11022](https://chartiq.kanbanize.com/ctrl_board/21/cards/11022/details)**

**Create logging export service example**
-Adds a microservice which listens to log messages, produces batches using a batch size or timeout and transmits them

**Description of testing**
-In its default form it tries to send the log messages to an invalid endpoint (as I don't have log collection endpoint to try this on at the moment), however the requests and batches look good.

### DIFF
--- a/src/services/loggingexport/loggingexport.html
+++ b/src/services/loggingexport/loggingexport.html
@@ -1,0 +1,10 @@
+<html>
+
+<head>
+    <script src="../../vendor.bundle.js"></script>
+</head>
+
+<body></body>
+<script src = "loggingexportService.js"></script>
+
+</html>
--- a/src/services/loggingexport/loggingexport.html
+++ b/src/services/loggingexport/loggingexport.html
@@ -1,8 +1,6 @@
 <html>
 
-<head>
-    <script src="../../vendor.bundle.js"></script>
-</head>
+<head></head>
 
 <body></body>
 <script src = "loggingexportService.js"></script>
--- a/src/services/loggingexport/loggingexportService.js
+++ b/src/services/loggingexport/loggingexportService.js
@@ -48,9 +48,33 @@ const CAPTURE_LOG_CATEGORIES = {
  */
 const SORT_MESSAGES = false;
 
+/**
+ * Message formatting function which converts each log message from Finsemble's format to the format you wish to transmit.
+ * Incoming log message format:
+ *		[										
+ *			{
+ *				"category": "system",				//Log message type: system, dev or perf
+ *				"logClientName": "Finsemble",		//The registered name of the logger instance
+ *				"logType": "Log",					//Log level: Error, Warning, Log, Info, Debug, Verbose
+ *				"logData": "[\"SERVICE LIFECYCLE: STATE CHANGE: Service initializing\",\"windowService\"]",
+ *													//JSON encoded array of message and data components of the log message
+ *													//N.B. maybe be prefixed by string "*** Logging Error: ""
+ *				"logTimestamp": 1544090028391.6226	//Log message timestamp for ordering use
+ *			}, 
+ *			{...},
+ *			...
+ *		]
+ */
+const FORMAT_MESSAGE = function (log_message) {
+	//TODO: add any necessary message format changes here
+
+	return log_message;
+};
+
 /** Where to transmit the logs to. */
 //TODO: Update to your logging endpoint
 const LOGGING_ENDPOINT = "http://somedomain.com/loggingendpoint";
+
 // #endregion
 
 /**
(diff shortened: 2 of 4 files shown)


kriswest:
@NinjaManatee Could you take a quick look at this logging export service example before I send it out to NatWest.

src/services/loggingexport/loggingexportService.js
```diff
@@ -0,0 +1,172 @@
...
+/**
+ * Log messages arrive in small batches from each logger and must be added to an export batch and 
+ * transmitted. 
+ * 
+ * @constructor
+ */
+function loggingexportService() {
+	const self = this;
+	let timeout = null;
+	let batchAllocSize = BATCH_SIZE + 10;     //add a small amount of leeway on batch size
```
NinjaManatee: I think this should be a constant.

src/services/loggingexport/loggingexportService.js
```diff
@@ -0,0 +1,172 @@
...
+	 * @private
+	 */
+	this.addToBatch = function (dataArr) {
+		if (dataArr) {
+			for (let m = 0; m < dataArr.length; m++) {
+				if (CAPTURE_LOG_CATEGORIES[dataArr[m].category] && CAPTURE_LOG_LEVELS[dataArr[m].logType]) {
+					//TODO: add any necessary message format changes here
+					logBatch[currBatchSize++] = dataArr[m];
+				} 
+				// else {
```
NinjaManatee: Would it be useful to keep this message as a `console.debug`? If not, remove the commented out code.

src/services/loggingexport/loggingexportService.js
```diff
@@ -0,0 +1,172 @@
+/**
+ * Simple log export service example which captures messages of specifed categories and levels
+ * and transmits them to a remote API in batch, once a certin batch size or timeout is reached.
+ * 
+ * The service will come up as soon as possible during finsemble startup - but will inevitably
+ * not be able to capture so of the earliest log messages.
+ * 
+ * Look for //TODO: comments for where to customize the service to your needs
```
NinjaManatee: Would it make sense to import a JS file with the necessary customizations?

src/services/loggingexport/loggingexportService.js
```diff
@@ -0,0 +1,172 @@
...
+	};
+
+	return this;
+};
+
+loggingexportService.prototype = new Finsemble.baseService({
+	startupDependencies: {
+		// Don't depend on any clients or services so that we start-up ASAP
+		services: [],
+		clients: []
```
NinjaManatee: Don't we need to depend on the RouterClient since we use it?

src/services/loggingexport/loggingexport.html
```diff
@@ -0,0 +1,10 @@
+<html>
+
+<head>
+    <script src="../../vendor.bundle.js"></script>
```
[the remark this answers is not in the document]
kriswest: I honestly have no idea... tried removing it, doesn't seem to change anything substantial, works just the same.

src/services/loggingexport/loggingexportService.js
```diff
@@ -0,0 +1,172 @@
...
+ * The service will come up as soon as possible during finsemble startup - but will inevitably
+ * not be able to capture so of the earliest log messages.
+ * 
+ * Look for //TODO: comments for where to customize the service to your needs
+ * 
+ * N.B. Be wary of adding too many Logger messages within this service as it will of course receive
+ * back to transmit onwards
+ */
+const Finsemble = require("@chartiq/finsemble");
+const RouterClient = Finsemble.Clients.RouterClient;
```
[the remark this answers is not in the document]
kriswest: No. Everything else uses it, so it has to be the first thing up I think.

src/services/loggingexport/loggingexportService.js
kriswest: Could do, but thats building an adapter (and fiddly) and really that should be loaded by the central logger so that we're only listening in one place AND we capture all messages (logger is up before loggerexportService). Core team know this and have put a few early thoughts into adapters, but nothing we can use right now.

src/services/loggingexport/loggingexportService.js
kriswest: No, I believe its created by routerClientInstance so its ready when referenced. Its the first thing up and all other clients depend on it, so depending on any other API client automatically covers it. If you don't need an API you don't need the router client.

kriswest:
I'll send this out to Natwest tomorrow. Should we start collecting a list of these recipes up somewhere? Guru?

src/services/loggingexport/loggingexport.html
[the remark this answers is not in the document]
NinjaManatee: I'll create a Twist conversation to check. I'll update the template when I get the answer.

NinjaManatee:
@kriswest The folder for this has everything that is needed for the example (as does the default workpsace example now). I think a Guru card that lists the examples and points to the folder with the set up instructions is a good first step. May be we should create a separate (private?) repo to hold the examples, so we don't have to worry about Finsemble versions.

kriswest:
https://app.getguru.com/card/ioEzjknT/Customer-Success-Recipes-Examples

PR #280, opened 2018-12-11
11168 tslint

erik-ciq:
**Resolves issue [11168](https://chartiq.kanbanize.com/ctrl_board/18/cards/11168/details)**

**Description of change**
- [x] Bring tslint into repo


carterbs:
@erik-ciq remember develop should be the base.

package.json
```diff
@@ -41,6 +43,7 @@
         "nocache": "^2.0.0"
     },
     "devDependencies": {
+        "@chartiq/tslint-config-chartiq-base": "^0.1.0",
```
carterbs: @erik-ciq the other PR also had the airbnb base included as a dependency. Why would it be needed in the other one and not in this one? Should it be in both or neither?

PR #282, opened 2018-12-12
Disable toolbar drag handle when docking is disabled

clausjs:
**Resolves issue [11175](https://chartiq.kanbanize.com/ctrl_board/18/cards/11175/details)**

**Description of change**
- Toolbar now checks its own dockable state and disabled drag handle when docking is disabled
- *REQUIRES*: https://github.com/ChartIQ/finsemble/pull/1041

**Description of testing**
-H Change 'dockable' to false in presentationComponents.json for 'Toolbar'. This will cause the drag handle to disappear and the toolbar will not be moveable

### DIFF
--- a/src-built-in/components/toolbar/src/dynamicToolbar.jsx
+++ b/src-built-in/components/toolbar/src/dynamicToolbar.jsx
@@ -44,7 +44,8 @@ export default class Toolbar extends React.Component {
 		super(props);
 		this.state = {
 			sections: ToolbarStore.getSectionsFromMenus(),
-			finWindow: fin.desktop.Window.getCurrent()
+			finWindow: fin.desktop.Window.getCurrent(),
+			canDrag: false
 		};
 		this.bindCorrectContext();
 	}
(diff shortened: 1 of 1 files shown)


mhmcclung:
for testing I verified maximize works correctly

src-built-in/components/toolbar/src/dynamicToolbar.jsx
```diff
@@ -62,6 +63,11 @@ export default class Toolbar extends React.Component {
 	componentDidMount() {
 		//console.log("this", this)
 		this.state.finWindow.bringToFront();
+		FSBL.Clients.ConfigClient.getValues({ field: "finsemble.components.Toolbar.window.dockable" }, (err, dockable) => {
+			this.setState({
+				canDrag: dockable
```
thorsent: @clausjs I think `canDrag` should probably be called `dockingEnabled`?

src-built-in/components/toolbar/src/dynamicToolbar.jsx
```diff
@@ -173,7 +179,7 @@ export default class Toolbar extends React.Component {
 		//console.log("Toolbar Render ");
 		if (!this.state.sections) return;
 		return (<FinsembleToolbar onDragStart={this.moveToolbar} onDragEnd={this.onPinDrag}>
-			<DragHandle />
+			{this.state.canDrag && <DragHandle />} {/*If this is not dockable, no need to show the drag handle */}
```
thorsent: @clausjs because down here I'd be asking myself, "what would cause canDrag to be false?"

PR #287, opened 2019-01-07
11483 Misleading Dialog Title

DanNicolai:
**Resolves issue [11483](https://chartiq.kanbanize.com/ctrl_board/18/cards/11483/details)**

**Description of change**
- Fixed a dialog title so that it's no longer ambiguous as to what happens when you name a New Workspace the same as an existing workspace.
- Commit is from Nicolai (my bad, Mike).

**Description of testing**
1. Create a New Workspace and name it the same as an existing workspace
2. A dialog will pop up with the title "New Workspace". (If it says "Overwrite Workspace?", then it didn't work.)

### DIFF
--- a/src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
+++ b/src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
@@ -581,7 +581,7 @@ Actions = {
 
 		let dialogParams = {
 			title: "New Workspace",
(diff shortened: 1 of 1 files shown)


DanNicolai:
Found an existing bug (reported by Daniel) in testing where new workspaces are not created when a custom named-workspace is used. 

https://chartiq.kanbanize.com/ctrl_board/18/cards/10467/details

Note that the correct behavior is demonstrated if you name the new workspace "Default Workspace". Otherwise, it will not work.

DanNicolai:
I'll place in the 'modified' language based on that suggestion. Thanks for pointing out the ambiguity.

PR #289, opened 2019-01-08
Don't enable allowRename if trying to rename active workspace, this a…

meredith-ciq:
**Resolves issue [11495](https://chartiq.kanbanize.com/ctrl_board/18/cards/11495/details)**

**Description of change**
-Removed ability to rename active workspace

**Description of testing**
-Attempted to rename current workspace in Preferences. Verified other workspaces could still be named and that it was possible to switch between two workspaces.


d4hines:
I've tested this and it works as expected.

My only comment is that there's a good bit of code duplication that makes this hard to read. But the whole module has this issue, and it can be refactored later - this for sure fixes the bug.

PR #292, opened 2019-01-09
11486 dockable window animate exception

meredith-ciq:
**Resolves issue [11486](https://chartiq.kanbanize.com/ctrl_board/18/cards/11486/details)**

**Description of change**
Using animate function added to DockableWindow. The existing function connected to has fewer parameters.
Requires: https://github.com/ChartIQ/finsemble/pull/1075

**Description of testing**
Make toolbar small and open and close search bar. Verify no exceptions occur.


d4hines:
Still, the behavior isn't quite right. It's supposed to grow with the search bar, but it's not yet.

PR #305, opened 2019-01-24
11789 e2o launch args

erik-ciq:
**Resolves issue [11789](https://chartiq.kanbanize.com/ctrl_board/18/cards/11789/details)**

**Description of change**
Add example of customizing launchE2O and launch args into gulpfile-extensions.js. Add envOrArg to taskMethods.

**Description of testing**
App still launches, e2o included the flags specified in gulpfile-extensions `launchE2o`

### DIFF
--- a/gulpfile-extensions-example.js
+++ b/gulpfile-extensions-example.js
@@ -14,7 +14,9 @@ module.exports = taskMethods => {
     "use strict";
 
     const gulp = require("gulp");
-
+	const ON_DEATH = require("death")({ debug: false });
+	const path = require('path');
+	const { exec } = require("child_process");
     /** -------------------------------------- PRE ----------------------------------------
 	  * You can use this section to run code that should execute *before* any gulp tasks run.
 	  * For instance, if you need to spin up a server or run a shell command, do it here.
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -437,6 +437,10 @@
 			logToTerminal.apply(this, arguments);
 		},
 
+		envOrArg: (...args) => {
```
NinjaManatee: Is it better to do this or just move the implementation of envOrArg here?

gulpfile-extensions-example.js
```diff
@@ -14,7 +14,9 @@ module.exports = taskMethods => {
     "use strict";
 
     const gulp = require("gulp");
-
+	const ON_DEATH = require("death")({ debug: false });
```
NinjaManatee: Whitespace is off

erik-ciq:
@NinjaManatee there is a lot of copy and pasted code, I'm also not a fan.

One alternate approach would be to do this via CLI flags / environment variables. So when a user launches finsemble-seed they could do something like: `npm run dev --auth-server-whitelist='*'`, or just set `auth-server-whitelist` as an environment variable. 

Another alternate approach is to create a new function like `launchE2OArgs` that returns an object of arguments to use when launching e2o. If we were to do that, we could eliminate almost all of the copy and paste code. 

I would be happy to update this PR with a better approach. I think it would be worth it.

gulpfile-extensions-example.js
carterbs: @erik-ciq please get with andy/confirm this comment is resolved.

gulpfile.js
carterbs: @erik-ciq please confirm with andy that he is satisfied and resolve this conversation.

carterbs:
QA notes....much of this may be due to me not knowing what I'm doing.

1. If I put anything after the manifest argument, the whole thing dies.
2. When it dies, it errors out because errorOutColor isn't a function.
3. If I put a nonsense param in the middle (`--twiddle yourThumbs`), the output is this:
```

0 -> C:\Users\BradC\git\finsemble-seed\node_modules\electron\dist\electron.exe                                                                                                                                                                                                                                                                                              1 -> index.js                                                                                                                                                                                                                                                                                                                                                               2 -> --auth-server-whitelist='*'                                                                                                                                                                                                                                                                                                                                            3 -> --remote-debugging-port=9090                                                                                                                                                                                                                                                                                                                                           4 -> NaN                                                                                                                                                                                                                                                                                                                                                                    5 -> --manifest                                                                                                                                                                                                                                                                                                                                                             6 -> http://localhost:3375/configs/openfin/manifest-local.json
--
```

I don't know if this is a bug or because it's trying to parse --twiddle as the port for the remote debugging port.

carterbs:
@erik-ciq take a look at my comments above. errorColorOut is the only thing that should probaby actually be fixed? Not sure if the other stuff is a bug.

erik-ciq:
@BradleyCarter went ahead and fixed the issue with errorColorOut, also I tested the `--twiddle yourThumbs` in the middle, that one launched for me once I fixed the errorColorOut issue. Everything seems to be working now.

gulpfile-extensions-example.js
```diff
@@ -27,6 +29,69 @@ module.exports = taskMethods => {
...
+		electronProcess.stdout.on("data", function (data) {
+			console.log(data.toString());
+		});
+
+		electronProcess.stderr.on("data", function (data) {
+			console.error("stderr:", data.toString());
+		});
+
+		electronProcess.on("close", function (code) {
+			console.log("child process exited with code " + code);
```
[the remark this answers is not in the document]
erik-ciq: I went ahead and updated the console log and removed the unnecessary commented out code.

gulpfile-extensions-example.js
```diff
@@ -27,6 +33,65 @@ module.exports = taskMethods => {
...
+
+			if (electronProcess) electronProcess.kill();
+
+			exec("taskkill /F /IM electron.* /T", (err, stdout, stderr) => {
+				// Only write the error to console if there is one and it is something other than process not found.
+				if (err && err !== 'The process "electron.*" not found.') {
+					console.error(errorOutColor(err));
+				}
+
+				if (watchClose) watchClose();
```
meredith-ciq: watchClose is throwing a not defined exception when I ctrl+c to kill finsemble. I'm not sure about regular shutdown because of the existing childprocess error, but this variable is not defined in this file.

erik-ciq:
Went ahead and updated the path to `electron.exe`, made sure that the `envOrArg.apply` is being returned properly, defined `watchClose` in `gulpfile-extensions-example.js`, and commented out `launchE2O` in `gulpfile-extensions-example.js`. Should be good now.

gulpfile.js
```diff
@@ -432,9 +432,7 @@
 			logToTerminal.apply(this, arguments);
```
meredith-ciq: @erik-ciq Can you make the same change as envOrArg here? It's only logging [object, object]. Otherwise these changes are great, thanks!

erik-ciq: @meredith-ciq went ahead and updated that.

gulpfile.js
```diff
@@ -433,9 +428,9 @@
 			}
 		},
 
-		logToTerminal: () => {
-			logToTerminal.apply(this, arguments);
-		},
+		logToTerminal: () => logToTerminal.apply(this, arguments),
```
meredith-ciq: @erik-ciq Sorry I just assumed the problem was the same as envOrArg without testing, but it's not. arguments is a reserved word with a lot of junk in it, but I don't think our message is inside. It needs to pass in args like envOrArg.
logToTerminal: (...args) => logToTerminal.apply(this, args),

erik-ciq: That makes sense. Updated to use args instead of arguments.

PR #310, opened 2019-02-01
11939 Added e2o version check

NinjaManatee:
**Resolves issue [11939](https://chartiq.kanbanize.com/ctrl_board/18/cards/11939/details)**

**Description of change**
- Added print of e2o version if `channelAdapter` is `e2o` and e2o is installed

**Description of testing**
- Clean checkout and install
- Run `npm run build` - builds successfully, no e2o output
- Change `channelAdapter` in `config/other/server-environment-startup.json` to `e2o`
- run `npm run build` - Throws error, suggests installing e2o
- Run `npm i @chartiq/e2o`
- Run `npm run build` - Builds successfully, output e2o version
- Change `channelAdapter` in `config/other/server-environment-startup.json` to `openfin`
- run `npm run build` - builds successfully, does not output e2o version

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -268,12 +268,16 @@
 			const CLI_VERSION = require(path.join(CLI_PATH, "package.json")).version;
 			const CONTROLS_PATH = path.join(__dirname, "node_modules", "@chartiq", "finsemble-react-controls");
 			const CONTROLS_VERSION = require(path.join(CONTROLS_PATH, "package.json")).version;
+			
+			// Check e2o version
 			const E2O_PATH = path.join(__dirname, "node_modules", "@chartiq", "e2o");
-			const E2O_VERSION = channelAdapter === "e2o" && fs.existsSync(E2O_PATH) ? require(path.join(E2O_PATH, "package.json")).version : undefined ;
-
-			if (channelAdapter === "e2o" && !fs.existsSync(E2O_PATH)) {
+			const E2O_PATH_EXISTS = fs.existsSync(E2O_PATH);
+			const USING_E2O = channelAdapter === "e2o";
+			if (USING_E2O && !E2O_PATH_EXISTS) {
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -268,6 +268,12 @@
 			const CLI_VERSION = require(path.join(CLI_PATH, "package.json")).version;
 			const CONTROLS_PATH = path.join(__dirname, "node_modules", "@chartiq", "finsemble-react-controls");
 			const CONTROLS_VERSION = require(path.join(CONTROLS_PATH, "package.json")).version;
+			const E2O_PATH = path.join(__dirname, "node_modules", "@chartiq", "e2o");
+			const E2O_VERSION = channelAdapter === "e2o" && fs.existsSync(E2O_PATH) ? require(path.join(E2O_PATH, "package.json")).version : undefined ;
```
garbee-ciq: I think this section has some are to improve legibility.

For example, checks are being duplicated here and down with the error. If the code was restructured a bit then it would be more expressive and increase legibility.

```
      const E2O_PATH = path.join(__dirname, "node_modules", "@chartiq", "e2o");
      const USING_E2O = channelAdapter === "e2o";
      const E2O_PATH_EXISTS = fs.existsSync(E2O_PATH) === true;

			if (USING_E2O === true && E2O_PATH_EXISTS === false) {
				throw "Cannot use e2o channelAdapter unless e2o optional dependency is installed. Please run npm i @chartiq/e2o";
			}

			const E2O_VERSION = require(path.join(E2O_PATH, "package.json")).version;
```

So, first we check if we are using e2o and if it exists in the filesystem up front. If we are using e2o and the path does not exist, throw the error. Then after, we know the path exists and that the adapter is in use, so we clean up the version grab to just get the version from the right file. This increases code legibility and has some slight micro-optimization performance gains by not having the engine run through loose type checks.

NinjaManatee:
@garbee-ciq I've made the changes you requested. Thanks!

PR #311, opened 2019-02-01
11567 listen for DOM resize event to resize the tabs.

carterbs:
**Resolves issue [11567](https://chartiq.kanbanize.com/ctrl_board/18/cards/11567/details)**

**Description of change**
-Previously we listened for bounds-set events on the finsemble windows.
-Moves to listening on the DOM's resize event.

This works because of the way the underlying container's events are abstracted away from the DOM. Even though the outer rectangle has been resized and an event has been received, the DOM may not have redrawn itself. Because tab sizing depends on the physical pixel size of the items in the DOM, this PR changes the listener from the finsemble window to the DOM window.

**Description of testing**
-Restored/maximized/resized a couple of stacks of various sizes.


src-built-in/components/windowTitleBar/src/components/center/TabRegion.jsx
```diff
@@ -18,7 +18,7 @@ let TAB_WIDTH = 300;
 const ICON_AREA = 29;
 const CLOSE_BUTTON_MARGIN = 22;
 const MINIMUM_TAB_SIZE = 100;
-
+import * as _throttle from "lodash.throttle";
```
garbee-ciq: We should just move this up with the other imports so they are all together. Minor code nit.

PR #312, opened 2019-02-05
11878 workspace dialogs

DanNicolai:
**Resolves issue [11878](https://chartiq.kanbanize.com/ctrl_board/18/cards/11878/details)**

**Description of change**
- Adjusted workspace dialog buttons so they reflect the actions that will be performed. (It's better UX than labeling them "Yes" or "No".)

**Description of testing**
1.  Dirty a workspace and perform a workspace switch. Check to see that saving dialog buttons read "Don't Save | Cancel | Save"

2. Choose "Save As" and name a workspace the same as another workspace. The prompting dialog's buttons should read "Cancel | Overwrite"

3. When deleting a workspace, the prompting dialog's buttons should read "Cancel | Delete"


carterbs:
Excellent. Looks good, tests well.

PR #315, opened 2019-02-07
11934 tabs bad state

meredith-ciq:
**Resolves issue [11934](https://chartiq.kanbanize.com/ctrl_board/18/cards/11934/details)**

**Dependent on Branch:** https://github.com/ChartIQ/finsemble/pull/1136

**Description of change**
- Added heartbeatListenerService as an example service on how to handle heartbeat events
- Listens for "NotResponding" heartbeat event and kicks window out of stack if it's in one.
- Listens for "crashed" heartbeat event and closes window

**Description of testing**
- See tests on https://github.com/ChartIQ/finsemble/pull/1136


src/services/heartbeatListener/heartbeatListenerService.js
```diff
@@ -0,0 +1,74 @@
...
+	/**
+	 * Responds to a HeartbeatTimeout Event
+	 * - Kicks windows out of a stack if they are not responding
+	 * - Kills crashed windows
+	 * @param {*} err 
+	 * @param {*} response 
+	 */
+	handleHeartbeat(err, response) {
+		//check if the window is in a stack, if so remove it, otherwise do nothing
+		if (response.data.type === "notResponding") {
```
garbee-ciq: Instead of an if series could we instead do a `switch` on the response type? This would allow us to have the other modes already provided with the default to return doing nothing. Making it clear the other modes are possible and where to modify them if needed in the future.

PR #318, opened 2019-02-12
[11740] e2o launch

rcsharp5:
**Resolves issue [11740](https://chartiq.kanbanize.com/ctrl_board/18/cards/11740/details#)**

This is a change that goes along with an e2o change for the index file. We now use `e2oLauncher` to run the electron version. The code was removed from the gulpfile for that. Use along with https://github.com/ChartIQ/e2o/pull/40. Everything should run as normal

**Description of testing**
Please see https://github.com/ChartIQ/e2o/pull/40


src/e2o/defaultStartup.js
```diff
@@ -0,0 +1,83 @@
+// Electron does not yet support "import"
+const { app } = require('electron');
+const e2o = require("@chartiq/e2o")
+//const FinsembleElectronUpdater = require('./src/startup/finsembleElectronUpdater');
+const log = require('electron-log');
+const superagent = require('superagent');
+const fs = require('fs');
+const MANIFEST_URL = "http://localhost:3375/configs/openfin/manifest-local.json"
```
carterbs: Should you get this from the application's config in case it's different from our default?

gulpfile.js
```diff
@@ -363,8 +364,12 @@
 			if (done) done();
 		},
 		launchE2O: done => {
+			let config = {
```
amaroks: I think this must be a `const`

src/e2o/defaultStartup.js
```diff
@@ -0,0 +1,79 @@
...
+const MANIFEST_URL = require('../../configs/other/server-environment-startup.json')[env.NODE_ENV || "development"].serverConfig
+const FEED_URL = 'http://localhost:3375/installers';
+// https://docs.microsoft.com/en-us/windows/desktop/shell/appids
+// Causes the multiple electron apps/windows to be grouped together under a single windows taskbar icon.
+app.setAppUserModelId('e2o');
+
+let updater = null;
+process.on('uncaughtException', (error) => {
+	log.warn(error);
+	log.error('error', error);
```
amaroks: Why 2 log messages for the same thing?

src/e2o/defaultStartup.js
```diff
@@ -0,0 +1,79 @@
...
+});
+// The installer will do some things here. If it needs to, it will return true and quit the app.(updates...)
+if (require('./src/startup/setupEvents')(app)) {
```
amaroks: I think loading module and executing code within `if()` is very ugly and considered a bad practice. Why not load at the top of file, and execute inside the if or before it and check the results?

I would only do lazy module loading when I'm not sure which module I need.

PR #319, opened 2019-02-14
12055 removed data-icon selector

DanNicolai:
**Resolves issue [12055](https://chartiq.kanbanize.com/ctrl_board/18/cards/12055/details)**

**Description of change**
- Removed data-icon selector from font-finance

**Description of testing**
- Create an adhoc component at the URL:  https://prd-p-4pl82v44zxvm.cloud.splunk.com/en-US/app/launcher/home
- Spawn the component
- Login using Andy's credentials:
user:  andy_chartiq
pass:  Reach closely creature agree1!
- When main dashboard loads, check to make sure the items in the top navigation Toolbar read "Splunk | Messages | Settings | Activity".  (i.e., they should not appear as jumbles of Finsemble icons).

### DIFF
--- a/assets/css/font-finance.css
+++ b/assets/css/font-finance.css
@@ -12,19 +12,6 @@
 
 }
 
-[data-icon]:before {
-  font-family: "font-finance" !important;
-  content: attr(data-icon);
-  font-style: normal !important;
-  font-weight: normal !important;
-  font-variant: normal !important;
-  text-transform: none !important;
-  speak: none;
-  line-height: 1;
-  -webkit-font-smoothing: antialiased;
-  -moz-osx-font-smoothing: grayscale;
-}
-
 [class^="ff-"]:before,
 [class*=" ff-"]:before {
   font-family: "font-finance" !important;


garbee-ciq:
Attempting to login with Andy's credentials fails for me. So, can't follow the instructs to verify the fix. After I get back from lunch if this isn't addressed here then I'll just make my own component that is hit by that selector and make sure everything jives right.

DanNicolai:
I was a little afraid that the site might flag Andy's username or something because of the logins from different computers. That might be what's happening. 

Also worth mentioning that it took me a couple tries to get in even with the correct password. Are you getting caught at the login saying you have incorrect credentials? Or are you able to get past that point?

garbee-ciq:
I just tried 5 or so times in a row. Each time I solved a captcha, then got told `Please check your username and password and try again.`

I can make this work without Splunk. All we need is a simple component that has a CSS selector with the same as what was removed. Add a target for it on the page, make sure we aren't touching it in the debugger.

DanNicolai:
Sure thing, lemme know how that works out.

PR #322, opened 2019-02-18
fix: toolbar being dockable

garbee-ciq:
**Resolves issue [12105](https://chartiq.kanbanize.com/ctrl_board/18/cards/12105/details)**

# Description fo change

- Moved `canGroup` turn off in the toolbar config

# Description of testing

1. Ran Finsemble with this config change
1. Opened Welcome Component
1. Moved WC to corner of toolbar
1. [ ] **Expect** the docking icon to not appear in the toolbar
1. [ ] **Expect** resizing the window from the right side to not resize the toolbar.

# Disclosure

I have no clue why this works at all. I'm not having any modifications jump out at me in either the seed or the framework code that would make this change needed.

### DIFF
--- a/configs/application/presentationComponents.json
+++ b/configs/application/presentationComponents.json
@@ -488,7 +488,6 @@
 				"minHeight": 39,
 				"minWidth": 300,
 				"docked": "top",
-				"canGroup": false,
 				"options": {
 					"autoShow": false,
 					"contextMenu": false,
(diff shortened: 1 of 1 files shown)


JimmyStevens:
This resolves 12105 and 12106 but introduces a bug where a component snapped (cascaded) and then resized does not snap to the edge of the monitor AND snaps to the toolbar but does not release. 

2u23dp

JimmyStevens:
canGroup: false // If canGroup is "false," the component can't be snapped to another component when floating.

From changelog 3.1.0 release.

PR #323, opened 2019-02-21
12053 change FDC3Server to appDirectoryEndpoint.

carterbs:
**Resolves issue [12053](https://chartiq.kanbanize.com/ctrl_board/18/cards/12053/details)**

**Description of change**
* Change FDC3Server to appDendpoint to better reflect what the config means.
* Change code to point to new config property.

**How to test**
1. Clone this repo: https://github.com/ChartIQ/fdc-appd, npm install, and npm run start.
2. Modify the value of `appDirectoryEndpoint` in `configs/application/config.json` to point to the local server. See `//appDendpoint` in the config for instructions.
3. Modify `config.json` in `src-built-in/toolbar` and change `App Launcher` (at the bottom) to `My Apps`.
4. Modify `presentationComponents.json` so that the component named `App Catalog` is named `Not the App Catalog`.
5. Modify `presentationComponents.json` so that the component named `App Catalog2` is named `App Catalog`. Now the application will point to the new app catalog. We're actively thinking about how to upgrade this component. Bear with us for now.
6. Start the seed. Make sure the local APPD server is running.
7. Click the apps menu, then launch the app catalog.
8. Verify that there is data.

### DIFF
--- a/src-built-in/components/appCatalog2/src/stores/storeActions.js
+++ b/src-built-in/components/appCatalog2/src/stores/storeActions.js
@@ -31,8 +31,8 @@ let FDC3Client;
 let appd;
 
 function initialize(done = Function.prototype) {
-	FSBL.Clients.ConfigClient.getValue({ field: "finsemble.FDC3Server" }, function (err, FDC3Server) {
-		FDC3Client = new FDC3({ url: FDC3Server });
+	FSBL.Clients.ConfigClient.getValue({ field: "finsemble.appDendpoint" }, function (err, appDendpoint) {
+		FDC3Client = new FDC3({ url: appDendpoint });
 		appd = new AppDirectory(FDC3Client);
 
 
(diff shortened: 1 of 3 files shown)


carterbs:
@rcsharp5 ian and I went back and forth over whether to capitalize the E. I hate the D and E being capitalized but it's not something i'm willing to die for. i'll change it out for you.

src-built-in/components/appCatalog2/src/stores/storeActions.js
```diff
@@ -31,8 +31,8 @@ let FDC3Client;
 let appd;
 
 function initialize(done = Function.prototype) {
-	FSBL.Clients.ConfigClient.getValue({ field: "finsemble.FDC3Server" }, function (err, FDC3Server) {
-		FDC3Client = new FDC3({ url: FDC3Server });
+	FSBL.Clients.ConfigClient.getValue({ field: "finsemble.appDirectoryEndpoint" }, function (err, appDirectoryEndpoint) {
```
amaroks: My thoughts, why not use async/await here to avoid having a whole logic in a callback function? There is going to be no way to write unit test in the future, if we wanted to.

Unless the ConfigClient's getValue does not return a promise, correct me if I'm wrong.

PR #327, opened 2019-02-25
chore: #id_no_card_for_this Update pull req template

carterbs:
chore: #id_no_card_for_this Update pull req template

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/no_card_for_this/details)

**Description of change**
* Make better PR template

**Description of testing**
1. Talked.
2. Made this PR.
- [x] Got sweet checkboxes.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,8 +1,13 @@
-**Resolves issue [####](https://chartiq.kanbanize.com/ctrl_board/18/cards/####/details)**
+fix:|feat:|BREAKING CHANGE: #id [CARD]
+
+[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/CARD/details)
 
 **Description of change**
--Change 1
--Change 2
+* Changes
 
 **Description of testing**
--How did you verify that your code satisfies the card's requirements?
+1. Step 1.
+2. Step 2.
+[x] Thing happened.
+[ ] Thing did not happen.
+3. Step 3
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,4 @@
-fix:|feat:|BREAKING CHANGE: #id [CARD]
+fix|feat|chore|docs|refactor|style|test: #id [CARD]
 
 [Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/CARD/details)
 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,8 +1,13 @@
...
 **Description of change**
--Change 1
--Change 2
+* Changes
 
 **Description of testing**
--How did you verify that your code satisfies the card's requirements?
+1. Step 1.
+2. Step 2.
+[x] Thing happened.
```
garbee-ciq: Ordered lists should all just be `1.` down. Markdown auto converts to the proper numbers when parsing.

Also for the checkboxes to be ordered right, you need them to be like `1. [ ] {content}`.

On the checkbox note as well, let's try to leave the template all empty up front. Ideally the description of testing should be copied and filled in when reviewing by reviewers. Assuming the person making the PR has followed the steps and gotten the expectations. otherwise they shouldn't be submitting the PR.

carterbs: Lines 9-10 are just there as examples. I think you and Erik were the only ones who knew how to make checkboxes. The template will continue to have text so that pertinent information isn't missed.

>Ideally the description of testing should be copied and filled in when reviewing by reviewers.

I actually like that the way I test is different from the way you test and the way Sidd tests (etc, etc, etc). If we're all doing the same things, I think we're wasting time. But that's a discussion for another time.

carterbs: > Ordered lists should all just be `1.` down. Markdown auto converts to the proper numbers when parsing.

I did not know this. I'll fix this.

garbee-ciq: > I actually like that the way I test is different from the way you test and the way Sidd tests (etc, etc, etc). If we're all doing the same things, I think we're wasting time. But that's a discussion for another time.

Testing differently is fine and we should hit different variations. However, the test steps provided should be the *baseline* to proving the fix works and addresses the given cases. Right now there are many ways to trigger different parts of the system, so expecting everyone to include all of it in a test pass is damaging. It isn't about being foolproof (without much more work into organizing knowledge of functionality), it's about having a solid baseline of the test flow.

Should something need to go into rework because a variation was now tested, then the test flow in the main PR description should be updated to reflect that, and future reviews would need to hit it as well.

carterbs: I think you're making valid points and your PR descriptions and code reviews are definitely being noticed. Keep beating the drum.

PR #346, opened 2019-03-21
hotfix: gulp crashing when e2o is not configured

garbee-ciq:
If e2o is not setup the build process should not break.
This *attempts* to require e2o, but does nothing if not found.
Only if the e2o code is being called do we then throw and error.

This prevents openfin only users from needing to setup e2o.

# How to verify the fix

1. Configure the seed to run using openfin.
1. Attempt to launch finsemble-seed attached to develop finsemble without this patch.
1. [ ] **Expect** the gulp process to fail right away since it can't find e2o exports.
1. Apply this patch
1. Run the seed again
1. [ ] **Expect** the gulp process to operate normally and launch Finsemble.
1. Setup e2o
1. Configure the seed to run on e2o
1. Run the seed again
1. [ ] **Expect** the gulp process to operate normally and launch Finsemble.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -19,8 +19,14 @@
 	const shell = require("shelljs");
 	const path = require("path");
 	const webpack = require("webpack");
-	const e2o = require("@chartiq/e2o/exports");
-	const e2oPackager = require("@chartiq/e2o/deploy/deploymentHelpers");
+	let e2o = null;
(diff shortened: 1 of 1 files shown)


garbee-ciq:
Since the electron adapter is going to be made required for launch, this PR won't be necessary.

PR #347, opened 2019-03-25
chore: #id 10865: Add NPM audit capability

mesner:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/10865/details)

**Meredith - Update 6/6/19:**
The decision was made to not use package-lock.json and block it's creation in the repos that weren't already doing so. Audits can be run every release and are still possible by temporarily removing the line blocking creation of the package-lock in npmrc and running npm i --package-lock-only. Package.json now requires specific versions so a lock file isn't needed. The existence of the lock file causes a lot of issues that npm either hasn't fixed or has only fixed in version 6.6 or later which is too new to be paired with node 8 via nvm. The main issues are that it is possible to end up in a state where npm i breaks our symlinks to the other repos everytime instead just occasionally and that there are bugs in npm < 6.6 that cause the lock file to change between npm install runs even with no changes. If we stay on node 8 it isn't possible to create a version of this file that won't cause git commit issues for all developers.

Also seed is a bit different because of all the nested dependencies to our other projects. Npm i does an audit run as well, but for some reason seed scans about 16k more files during npm i than npm audit. Npm audit is the only way to see the issues, so there is one remaining high alert issue that only shows up on npm i. I have no idea what it is so I can't fix it, but when npm audit is run there are no high alert issues.

**Description of change**
* Added package-lock.json to repository
* Converted npm dependencies to asterisk-based for clarity
* Froze minor versions of dependencies to most recent versions that matched prior tilde-based criteria

**Description of testing**
1. Basic startup functionality


erik-ciq:
Updated against `planned/3.10.0`. Everything still starts okay. Ready for review again.

PR #348, opened 2019-03-25
12614 preferences component always on top

clausjs:
**Resolves issue [12614](https://chartiq.kanbanize.com/ctrl_board/18/cards/12614/details)**

Working example: https://www.dropbox.com/s/pzd1s64ipng56wg/QTq79hvuoT.mp4?dl=0
**(You'll notice at the end, after the dialogs are used/cancelled (tested both) the welcome component cannot come in front of the preferences component)**

**Description of change**
- Adds a state variable to the Workspace component which is the status of the window's alwaysOnTop option
- Adds a component level function which will call the window client to set the alwaysOnTop option for the preferences component and set the internal state's copy of the same variable
- When the window is set to not be on top, an event listener is added that resets the state when focus is received back on the Workspace window (this handles when the file dialog is closed, regardless of whether something was chosen)
- Adds a check on initial component initiation which will ignore all calls to change alwaysOnTop if the preferences component's alwaysOnTop is set to false on initialization

**Description of testing**
- If `UserPreferences.window.options.alwaysOnTop` is set to true in `presentationComponents.json` then open the Preferences component. The window should be on top of everything else. When Workspaces > Import OR Export is clicked, the file dialog should open on top of the preferences component (or if it open to the side of it, should be able to drag it on top of the component). When the file dialog is closed, whether a file was chosen/saved or not, the Preferences component should return to being alwaysOnTop.
- [x] Preferences component was always on top unless file manager dialog was open
- [ ] Preferences component was not always on top

- If `UserPreferences.window.options.alwaysOnTop` is set to false in `presentationComponents.json` then none of this will happen, and preferences component can always have windows on top of it.
- [ ] Preferences component was not always on top

### DIFF
--- a/src-built-in/components/userPreferences/src/components/content/Workspaces.jsx
+++ b/src-built-in/components/userPreferences/src/components/content/Workspaces.jsx
@@ -408,7 +408,7 @@ export default class Workspaces extends React.Component {
 					name: activeName,
 					force: true
 				}, (err, response) => {
-						doExport();
+					doExport();
 				});
 			}
 			doExport();
--- a/src-built-in/components/userPreferences/src/components/content/Workspaces.jsx
+++ b/src-built-in/components/userPreferences/src/components/content/Workspaces.jsx
@@ -326,7 +326,7 @@ export default class Workspaces extends React.Component {
 		//Set alwaysOnTop to false and add an event listener on the window. When focus is regained
 		//then reset always on top
 		this.changeAlwaysOnTop(false);
-		finsembleWindow.addEventListener('focused', this.preferencesFocused, false);
+		finsembleWindow.addEventListener('focused', this.preferencesFocused);
 		let inputElement = document.getElementById("file-input");
 		inputElement.addEventListener("change", importWorkspace, false);
 		inputElement.click();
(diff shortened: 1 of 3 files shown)


src-built-in/components/userPreferences/src/components/content/Workspaces.jsx
```diff
@@ -303,16 +323,32 @@ export default class Workspaces extends React.Component {
...
+	preferencesFocused() {
+		finsembleWindow.removeEventListener('focused', this.preferencesFocused);
+		this.changeAlwaysOnTop(true);
 	}
 
 	importWorkspace(evt) {
 		let inputElement = document.getElementById("file-input");
 		inputElement.removeEventListener("change", importWorkspace);
 		var files = evt.target.files; // FileList object
+		var self = this;
```
meredith-ciq: Not used

src-built-in/components/userPreferences/src/components/content/Workspaces.jsx
```diff
@@ -108,6 +110,24 @@ export default class Workspaces extends React.Component {
 		UserPreferencesStore.addListener({ field: 'WorkspaceList' }, this.setWorkspaceList)
 	}
 
+	/**
+	 * Sets the windowOptions.alwaysOnTop to the value supplied. The preferences component should
+	 * not be always on top when a file dialog is open.
+	 * @param {boolean} alwaysOnTop The value to set finsembleWindow.options.alwaysOnTop
+	 */
+	changeAlwaysOnTop(alwaysOnTop) {
```
meredith-ciq: If this is only used for preferences or even for dialogs, I'd suggest renaming this. It's a little confusing why this is called in the code with a generic name.

PR #361, opened 2019-04-19
fix: #id [13097] Moved Electron existence check to launchElectron

NinjaManatee:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/13097/details)

**Description of change**
* Moved Electron existence check to launchElectron 

**Description of testing**
1. Run `git clean -xfd` and make sure _node_modules_ is deleted
1. Remove `"@chartiq/finsemble-electron-adapter"` entry from `optionalDependencies` in _package.json_
1. Run `npm i`
1. Run `npm run build`
1. [x] Finsemble builds without error
1. Run `npm run dev`
1. [x] Finsemble fails to run with the following error:
    `Cannot use electron container unless finsemble-electron-adapter optional dependency is installed. Please run npm i @chartiq/finsemble-electron-adapter`
1. Revert change to _package.json_
1. Run `git clean -xfd`
1. Run `npm i`
1. Run `npm run build`
1. [x] Finsemble builds without errors
1. Run `npm run dev`
1. [x] Finsemble runs without errors

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -289,12 +289,6 @@
 			const CONTROLS_PATH = path.join(__dirname, "node_modules", "@chartiq", "finsemble-react-controls");
 			const CONTROLS_VERSION = require(path.join(CONTROLS_PATH, "package.json")).version;
 
-			// Check electron adapter version
-			const USING_ELECTRON = container === "electron";
-			if (USING_ELECTRON && !FEA_PATH_EXISTS) {
-				throw "Cannot use electron container unless finsemble-electron-adapter optional dependency is installed. Please run npm i @chartiq/finsemble-electron-adapter";
-			}
-
 			// Check version before require so optionalDependency can stay optional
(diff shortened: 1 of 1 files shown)


carterbs:
@NinjaManatee 

Why is this being done? The error only throws if you are using electron AND you don't have the package installed. If your config is set to use openfin, the error will not be thrown. By moving the logic check, you're not getting rid of the error, just making it happen in a different spot. 

I'm fine with that change, but it doesn't seem to address the underlying concern (e.g., "Build shouldn't fail for an optional dependency"). Because the default container is electron, if you don't install FEA, you will get this error. I think it's a helpful error and should be thrown. 

Let me know what I'm missing. I'm fine with merging this as is.

carterbs:
I think we figured it out -- you're doing this so that you can build the seed and sales-demo on CI. You're not running it. You're just building it. That check was causing your build to fail.

PR #371, opened 2019-05-03
fix: #id 12914 fix double-click not restoring when maximized

meredith-ciq:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/12914/details)

**Description of change**
* There are two places calling maximize when receiving the openfin maximized event. This often is okay, but occasionally the extra call to maximize triggers a maximize call when we're trying to restore. This is a race condition and happens fairly infrequently. Removes the extra listener to maximized so the window service can be the only place that listens for these events.

**Description of testing**
You can try to reproduce the bug before the fix, but it happens very infrequently. That makes it hard to test that this fix works, but we can have some confidence if everyone tries several times and the removal of this listener causes no issues. Test on openfin, os doesn't matter as this bug is not windows 7 specific, though if you have a vm it seems to be easier to hit the problem.
1. Open two welcome components and group them.
1. [x] On the same window as the toolbar, doubleclick one of their titlebars to maximize. That window should maximize under the titlebar
1. [x] Double click the maximized window again to restore. The window should restore back to where it was before in the group.
1. Try this several times, double-click on both windows. If you hit this bug at anytime please export a copy of the logs for me to look at.


erik-ciq:
I wasn't able to reproduce the bug. I did detect a separate issue when testing. The separate issue can be handled in a different card.

1. Floating toolbar
2. Maximize docked welcome component
3. Notice it leaves space for a toolbar as if it wasn't floating
4. Unmaximize, notice that it is offset from original position ~ the height of the taskbar.

Begin state:
image

Maximized:

image

After unmaximize:

image

meredith-ciq:
That's written up here: https://chartiq.kanbanize.com/ctrl_board/48/cards/13377/details/ It's electron only and I don't get it on win 7.

PR #373, opened 2019-05-03
fix: #id 13417 Don't check isShowing when menuWindow is null

clausjs:
fix: #id [13417]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/13417/details)

**Description of change**
* Add a check before handleClose calls menuWindow.isShowing to ensure menuWindow isn't null before checking

**Description of testing**
1. Start sales demo
1. Login
1. Open Central Logger
1. [x] Central logger should not have any errors pertaining to `Cannot read property 'isShowing' of null`


src-built-in/components/toolbar/stores/searchStore.js
```diff
@@ -174,24 +174,26 @@ var Actions = {
...
 	handleClose(e) {
-		menuWindow.isShowing(function (err, showing) {
-			if (showing) {
-				console.log("close a window")
-				if (!e && cachedBounds) {
-					finsembleWindow.animate({ transitions: { size: { duration: 150, width: cachedBounds.width } } }, () => {
-						cachedBounds = null;
-					});
+		// Don't call isShowing if menuWindow is not set yet
+		if (menuWindow) {
```
garbee-ciq: Could we not write this as `if (! menuWindow){ return; }` so the commit doesn't touch things unnecessarily?

PR #383, opened 2019-05-15
12215 toolbar hotkey to config

WatsonCIQ:
fix: #id [12215]

https://chartiq.kanbanize.com/ctrl_board/21/cards/12215/details/
~~https://chartiq.kanbanize.com/ctrl_board/21/cards/11551/details/~~
~~https://chartiq.kanbanize.com/ctrl_board/21/cards/11848/details/~~
~~https://chartiq.kanbanize.com/ctrl_board/21/cards/12268/details/~~
~~https://chartiq.kanbanize.com/ctrl_board/21/cards/13493/details/~~

**Moved toolbar hotkeys into the toolbar config file with default, a user can enable or disable the hotkey from the config file.**
 ~~_Sorry for all the fixes in one PR but they are all related and small_~~_Since starting this most of the above tickets have been completed elsewhere by other people_

* The toolbar config is now referred to as the toolbar config, not menu config in toolbar store. This is due to the fact it houses both hotkeys and the menus. I have added backward compatibility so that developers can still reference the menu items with the same variable.
* Added hotkey section to the toolbar config file. If a hotkey name is not included in the list it will use the default hotkey combination.
~~* Fixed the 'No Results Found' notice~~
~~* Search now comes to the front when using the hotkey **ctrl+alt+f**~~
~~* Search now has focus and can start typing when using the hotkey **ctrl+alt+f**~~

**Description of testing for hotkey config**
1. Change one or more hotkey combinations in the toolbar config file.
1. Run dev Finsemble and test the hotkey.
1. [x] Hotkeys act as expected and if an action name and combination are not referenced in the config then a default is applied.
1. [ ] Hotkeys are broken / not working...


src-built-in/components/toolbar/components/Search.jsx
```diff
@@ -106,7 +106,7 @@ export default class Search extends React.Component {
 	componentDidUpdate() {
 		if (this.state.hotkeySet) {
 			FSBL.Clients.WindowClient.finWindow.focus(() => {
-				this.searchContainer.current.focus();
+				this.searchContainer.current.querySelector('#searchInput').focus()
```
joel-ciq: Is there a way to follow Reacts way of accessing components vs. selecting dom elements directly? Maybe it's really hard in the current framework...

src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -251,25 +244,69 @@ class _ToolbarStore {
...
+	/**
+ * Load the hotkeys from the config.json. If there are no items in config.json, hotkeys use the defaults in the current file.
+ *
+ *
+ * @param {any} done
+ * @memberof _ToolbarStore
+ */
+	loadHotkeysFromConfig(done) {
+
+		// destructure toolbarConfig.hotkeys and provide a default fallback value for each hotkey.
```
joel-ciq: Nicely done

src-built-in/components/toolbar/components/Search.jsx
```diff
@@ -136,8 +136,15 @@ export default class Search extends React.Component {
 	setActive(err, data) {
 		this.setState({ active: data.value })
 	}
-	hotKeyActive() {
-		this.setState({ active: true, hotkeySet: true })
+	hotKeyActive(...args) {
+		// the first argument seems to be empty but the second value returns
```
NinjaManatee: The first argument is likely an error object, and will be null or undefined if there isn't an error. If there is an error, it should be logged.

src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -251,25 +244,69 @@ class _ToolbarStore {
...
+			showToolbar,
+			() => {
+				this.Store.setValue({ field: "searchActive", value: false });
+				this.showToolbarAtFront();
+			});
+
+		// hide the toolbar
+		FSBL.Clients.HotkeyClient.addGlobalHotkey(
+			hideToolbar,
+			() => {
```
NinjaManatee: Is the anonymous function necessary here? Can't it just be:
```
FSBL.Clients.HotkeyClient.addGlobalHotkey(hideToolbar, this.hideToolbar);
```
I think this is true for all of them except showToolbar and showSearch.

carterbs:
@WatsonCIQ how did you make a draft PR?

@NinjaManatee Are you guys planning on moving this forward? It solves an impressive number of issues.

WatsonCIQ:
> @WatsonCIQ how did you make a draft PR?
@carterbs check it out here: https://github.blog/2019-02-14-introducing-draft-pull-requests/
> @NinjaManatee Are you guys planning on moving this forward? It solves an impressive number of issues.
@carterbs I think many of those issues have since been resolved 🏁, I am going to go back through and check which have been resolved and check them off the list above✅. 
I have started to pull those changes to make it easier to merge 👓. 
The last thing I need to do is to modify my code slightly as my code as it does not add backwards compatibility 😳. 
I should be done with the changes soon. I will also resolve the changes Andy has requested.
👍

src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -246,33 +245,166 @@ class _ToolbarStore {
...
-	 */
+	* Load the hotkeys from the config.json. If there are no items in config.json, hotkeys use the defaults in the current file.
+  *
+  *
+	* @param {any} done
+	* @memberof _ToolbarStore
+	*/
+	setupHotkeys(done) {
+		// defaults for the hotkeys
+		const defaultHotkeys = [
```
joel-ciq: How about:
A) Moving the defaults array to a separate file that can be used in multiple places? 
B) Making the action strings constants.

src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -246,33 +245,166 @@ class _ToolbarStore {
...
+			{
+				name: "showToolbar",
+				"keyCombination": [
+					"ctrl",
+					"alt",
+					"t"
+				],
+				action: 'SHOW_TOOLBAR',
```
joel-ciq: How about making the action strings constants in a separate file?

src-built-in/components/toolbar/stores/toolbarStore.js
WatsonCIQ: I like this idea. 
I think moving forward it would be a great idea as clashes in hotkeys are not ideal anyway!

src-built-in/components/toolbar/config.json
```diff
@@ -1,56 +1,110 @@
...
-        "menuType": "App Launcher"
-    },
-    {
-        "align": "left",
-        "type": "seperator"
-    }
-]
\ No newline at end of file
+{
+    "menus": [
```
NinjaManatee: 💯

src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -3,18 +3,25 @@
...
 import async from "async";
-import * as menuConfig from '../config.json';
+import * as toolbarConfig from '../config.json';
 import * as storeExports from "../stores/searchStore";
 
 import { Actions as SearchActions } from "./searchStore"
 
+// this refers to the new menu section in the toolbar, to add backward compatibility
+// if this structure does not exist then just default to using the old array structure
+const menuConfig = 'menus' in toolbarConfig ? toolbarConfig.menus : toolbarConfig
```
NinjaManatee: Should you use ' or "?

WatsonCIQ: " for consistency, thanks for spotting this @NinjaManatee 👍

src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -246,33 +245,169 @@ class _ToolbarStore {
...
+
+		/**
+		 * updateDefaultHotkeysFromConfig
+		 *
+		 * This function takes the hotkeys from the config file and updates / merges the default hotkeys with it's values.
+		 * @param {Array} configHotkeys
+		 * @param {Array} defaultHotkeys
+		 */
+		const overrideDefaultHotkeysWithConfigHotkeys = (configHotkeys, defaultHotkeys) =>
+			defaultHotkeys.map(defaultHotkey => {
```
NinjaManatee: This should probably be:
```suggestion
			defaultHotkeys.forEach(defaultHotkey => {
```

src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -246,33 +245,169 @@ class _ToolbarStore {
...
+
+				// This checks to see if there is an object inside the default hotkey array that matches
+				// one from the configHotkeys - validation check
+				const config = configHotkeys
+					.find(configHotkey => defaultHotkey.name === configHotkey.name)
+
+				// return one object by merging the two objects into one object.
+				// If there is a value already set in both objects then the right-hand side (configHotkeys)
+				// overwrites the left (hotkeys)
+				return { ...defaultHotkey, ...config }
```
NinjaManatee: Wouldn't this make it impossible to remove a default hotkey?

src-built-in/components/toolbar/stores/toolbarStore.js
WatsonCIQ: It's worth a conversation, I'm totally open to ideas on this one.

NinjaManatee: I thought you were pulling the defaults from file. Is that not the case?

src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -246,33 +245,169 @@ class _ToolbarStore {
 	}
 
 	/**
-	 *
-	 *
-	 *
-	 */
+	* Load the hotkeys from the config.json. If there are no items in config.json, hotkeys use the defaults in the current file.
+  *
```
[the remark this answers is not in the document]
WatsonCIQ: This seems to be fine when you go to edit the file and in VS code. I'm guessing it is just the review view?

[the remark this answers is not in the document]
NinjaManatee: Try changing the size of the tabs. I think one line has tabs and the other has spaces.

src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -246,33 +245,169 @@ class _ToolbarStore {
...
+		 * This function takes the hotkeys from the config file and updates / merges the default hotkeys with it's values.
+		 * @param {Array} configHotkeys
+		 * @param {Array} defaultHotkeys
+		 */
+		const overrideDefaultHotkeysWithConfigHotkeys = (configHotkeys, defaultHotkeys) =>
+			defaultHotkeys.map(defaultHotkey => {
+
+				// This checks to see if there is an object inside the default hotkey array that matches
+				// one from the configHotkeys - validation check
+				const config = configHotkeys
```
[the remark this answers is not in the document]
WatsonCIQ: Unfortunately not as configHotkeys is an array of objects, defaultHotkey is an object. It looks like this [{},{},{}] where configHotkeys is the array and defaultHotkey is an object.
Again maybe the comments are not very clear?

src-built-in/components/toolbar/stores/toolbarStore.js
WatsonCIQ: I don't think so as this is only just merging the objects (no hotkeys are set yet).
The hotkeys get set in a function lower down, does this sound correct? 
I have also just changed the block comment to make this a bit clearer.

src-built-in/components/toolbar/stores/toolbarStore.js
WatsonCIQ: Can you elaborate on the change from map to forEach? Is it a functional reason or is it not clear? 🙂

src-built-in/components/toolbar/stores/toolbarStore.js
WatsonCIQ: Not defaults, just config overrides. The reason was not to break existing clients code. Thinking about it now I'm sure we could get rid of these. This would get rid of some of the other code below too. I need to revisit and reason to why I kept defaults.

src-built-in/components/toolbar/stores/toolbarStore.js
amaroks: @NinjaManatee ?

NinjaManatee: @WatsonCIQ @amaroks `map` is for transforming one array into another and `forEach` is for looping over an array of objects. Since you aren't capturing the returned result of the mapping, I think a `forEach` is more correct. 

Upon reading this code again, I see where my confusion is. `overrideDefaultHotkeysWithConfigHotkeys` is a one line function, whose return is inferred to be the result of the mapping. For clarity, I think it would be good to add {} around the function and explicitly add the return.

PR #387, opened 2019-05-21
id: #13493 fix search bar highlighting text on selection

zona-ciq:
fix: #id [13493]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/13493/details/)

**Description of change**
* Remove the range (highlight) of the search text when user click away
* Remove duplicated code

**Description of testing**
1. Run Finsemble seed
1. Type something in the search bar in the tool bar
1. Select a search result
1. Click into the search bar again and select the search result
1. [x] We can no longer see the highlighted search text

### DIFF
--- a/src-built-in/components/toolbar/stores/searchStore.js
+++ b/src-built-in/components/toolbar/stores/searchStore.js
@@ -50,7 +50,7 @@ var Actions = {
 	},
 	setFocus(bool, target) {
 		focus = bool;
-		if (!menuWindow) return;
+		if (!menuWindow) return Actions.handleClose();
 		if (bool) {
 			if (window.outerWidth < 400) {
 				finsembleWindow.getBounds((err, bounds) => {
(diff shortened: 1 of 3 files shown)


src-built-in/components/toolbar/stores/searchStore.js
```diff
@@ -76,11 +76,11 @@ var Actions = {
 				Actions.positionSearchResults();
 			});
 
+		} else {
+			var sel = window.getSelection();
```
amaroks:
```suggestion
		
```

src-built-in/components/toolbar/stores/searchStore.js
```diff
@@ -76,11 +76,11 @@ var Actions = {
...
+			sel.removeAllRanges();
```
amaroks:
```suggestion
			window.getSelection().removeAllRanges();
```

PR #390, opened 2019-05-23
fix: #id 13967 Fix search menu hide behavior on enter

carterbs:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/13967/details)

* Previously, if one typed in a search, pressed the down arrow, and
pressed enter, the search menu would reappear after launching the component.

This is fixed.

**Description of change**
* Changes

**Description of testing**
1. Type into the seach box 'we'.
1. Press the down arrow.
1. Hit enter.

Observe that the search menu hides and does not reopen.

### DIFF
--- a/src-built-in/components/toolbar/components/Search.jsx
+++ b/src-built-in/components/toolbar/components/Search.jsx
@@ -101,7 +101,14 @@ export default class Search extends React.Component {
 	}
 
 	textChangeDebounced(event) {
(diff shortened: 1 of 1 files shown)


carterbs:
Spoke with garbee. His apprehension was that the textbox stays open ocassionally and the card said that the menu should close and the textbox should close. I didn't see the textbox issue in the card. We have other cards for fixing the textbox. Will move this to RTM.

PR #391, opened 2019-05-27
fix: #13977 Edit workspaces name in preferences

amaroks:
fix: #id13977

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/13977/details)

**Description of change**
* Removed unnecessary `this.cancelEdit()` from `setWorkspaceList()`

**Description of testing**

- Checkout this branch 
- Launch finsemble `npm run dev`
- Open user preferences > workspaces

- [ ] You can rename workspaces by clicking on the pen icon

### DIFF
--- a/src-built-in/components/userPreferences/src/components/content/Workspaces.jsx
+++ b/src-built-in/components/userPreferences/src/components/content/Workspaces.jsx
@@ -104,7 +104,6 @@ export default class Workspaces extends React.Component {
 		this.setState({
 			workspaceList: data.value
 		});
-		this.cancelEdit();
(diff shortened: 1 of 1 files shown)


garbee-ciq:
Tested with the current 3.9.0 branches for FEA and Finsemble. Works as expected and allows editing again, yay!

I'm concerned as to why this was working before though. Since this line has been there since at least 2.2.0 last year and it is only just now breaking on us. But, that's something that shouldn't really hold anything up since this is purely a UX thing and this is fixing it.

PR #413, opened 2019-06-21
Fix #14723 built in defaultAuthentication initialization.

gerard-ciq:
Fix built in defaultAuthentication initialisation.

fix: #id [14723]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/48/cards/14723/details/)

**Description of change**
1. Built in auth component to use FSBL on ready
2. Can now submit username and password using keypad enter too
3. editors no longer add trailing space to .json files
4. Login Window is now correct size.

**Description of testing**
Add to application/config.json
```
    "authentication": {
		"startup": {
			"adapter": "PASSWORD",
			"component": "defaultAuthentication"
		}
	}
````
Login window shows

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,3 +11,6 @@ indent_size = 2
 indent_style = tab
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*.json]
+insert_final_newline = false
--- a/configs/application/components.json
+++ b/configs/application/components.json
@@ -155,7 +155,7 @@
 				"top": "center",
 				"left": "center",
 				"width": 358,
-				"height": 458,
+				"height": 658,
 				"ephemeral": true,
 				"options": {
 					"autoShow": false
--- a/src-built-in/components/authentication/authentication.js
+++ b/src-built-in/components/authentication/authentication.js
@@ -3,12 +3,15 @@ var inLogin = true;
  * 			SAMPLE AUTHENTICATION
  ************************************************/
 //On ready, check to see if the user has a valid session
-FSBL.Clients.RouterClient.onReady(() => {
-    checkAuthorizationStatus();
-});
+if (window.FSBL && FSBL.addEventListener) {
+	FSBL.addEventListener("onReady", checkAuthorizationStatus);
+} else {
+	window.addEventListener("FSBLReady", checkAuthorizationStatus)
+}
+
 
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: .editorconfig [hunk 1]
```diff
@@ -10,4 +10,7 @@ end_of_line = lf
 indent_size = 2
 indent_style = tab
 insert_final_newline = true
-trim_trailing_whitespace = true
\ No newline at end of file
+trim_trailing_whitespace = true
+
+[*.json]
+insert_final_newline = false
```
gerard-ciq: I saw in a different review that someone was asked to remove the trailing space from the config file. Is this something we do? If so I put this in here so editors don't add one by default.

If it's not a thing I will remove this :)

NinjaManatee: I don't know whether that is a thing, but I'm leery of adding changes that are related to the ticket.

amaroks:
I think this was supposed to be merged into `planned/3.9.0` instead of develop.

gerard-ciq:
@amaroks 🤷‍♂ I honestly don't know how the team manages versioning and changes between branches. Should I change it?

kriswest:
@gerard-iq and @amaroks, should that be planned/3.10 now?

More generally, we're somewhat further from the releases than you guys so
it probably make sense for the reviewer on [Review and merge] tickets from
CS to set the base... We don't actually know ahead of time which branch
you'll be merging into...

CC: Brent and Sidd
K

Kris West
Senior Solutions Architect
p: +1 800 821 8147  m: +44 (0)7715104292
w: www.chartiq.com  e: kris@chartiq.com

.editorconfig
gerard-ciq: You're not sure what is a thing? The convention that we remove spaces at the end of config files? If that's the case that makes me happy... intelliJ adds a spaces at the end if it's true. If it's not a convention we have, that makes me happy and I'll remove it and I'll be less afraid of checking in whitespace only files by mistake.

amaroks:
@sgd2z I changed the base branch to `planned/3.10.0` let me know if this is not ok.

.editorconfig
sgd2z: if this thing prevents that constant package.json final newline, it would be great!

gerard-ciq: It does yes :)

PR #414, opened 2019-06-21
fix: #id 14688 app cant unpin when name changes

carterbs:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/14688/details)

**Description of change**
This PR makes it so that pins are renamed when the underlying component has a display name change. Additionally, if the component is removed from components.json, the pin will be excluded from the toolbar.

This PR also fixes tsconfig and ts-loader issue I had while working.

**Description of testing**
1. Load Finsemble on planned/3.10.0.  Pin a workspace and every component that is loaded. Switch to this branch and run `npm install`.
2. In components.json, change the component:displayName of Welcome Component to some other name.
3. In components.json, remove the Notepad entry.
4. Restart Finsemble.
5. Confirm that the pin for the Welcome Component has changed its name to whatever you named it to.
6. Confirm that the pin for Notepad is gone.
7. Unpin Welcome Component.
8. Restart Finsemble.
9. Confirm that the pin for Welcome Component is gone.
10. Add Notepad back.
11. Pin it.
12. Confirm that Notepad is pinned.
13. Restart.
14. Confirm that everything is the same.

### DIFF
--- a/src-built-in/components/appLauncher/src/stores/appLauncherStore.js
+++ b/src-built-in/components/appLauncher/src/stores/appLauncherStore.js
@@ -223,7 +223,6 @@ var Actions = {
 		}
 
 		//No dots allowed in keys in the global store, so escaping dots
-		var componentTypeDotRemove = componentType.replace(/[.]/g, "^DOT^");
 
 		var pins = appLauncherStore.getValue("pins");
 		let params = { addToWorkspace: true, monitor: "mine" };
@@ -259,6 +258,8 @@ var Actions = {
 		if (!wasPinned) {
 			pins.push(thePin);
 		}
+
+		var componentTypeDotRemove = displayName.replace(/[.]/g, "^DOT^");
 
 		if (wasPinned) {
 			ToolbarStore.removeValue({ field: "pins." + componentTypeDotRemove });
--- a/src-built-in/components/toolbar/modules/pinManager.ts
+++ b/src-built-in/components/toolbar/modules/pinManager.ts
@@ -1,0 +1,92 @@
+/**
+ * This file handles component displayname changes and cases where
+ * components are no longer registered with finsemble.
+ * Given an array of pins and a component list, it'll do the work for you.
+ */
+declare type Pin = {
+	component?: string,
+	fontIcon: string,
+	icon: string,
+	index: number,
+	label: string,
+	params: {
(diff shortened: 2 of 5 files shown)


src-built-in/components/appLauncher/src/stores/appLauncherStore.js
```diff
@@ -260,6 +259,8 @@ var Actions = {
 			pins.push(thePin);
 		}
 
+		var componentTypeDotRemove = displayName.replace(/[.]/g, "^DOT^");
```
carterbs: Changed from componentType to displayName. This allows pinning to work with the display name, not the component type.

src-built-in/components/toolbar/modules/pinManager.ts
```diff
@@ -0,0 +1,92 @@
...
+export class PinManager {
+	/**
+	 * Converts the array of pins into an object where the key is the label
+	 * of the pin
+	 *
+	 * @param {Pin[]} arr
+	 * @returns any
+	 */
+	convertPinArrayToObject(arr: Pin[]): any {
+		let obj: any = {};
```
amaroks:
```suggestion
		const obj: any = {};
```

src-built-in/components/toolbar/modules/pinManager.ts
```diff
@@ -0,0 +1,92 @@
...
+		arr.forEach((el, i) => {
+			if (el) {
+				let key = el.label;
```
amaroks:
```suggestion
				const key = el.label;
```

src-built-in/components/toolbar/modules/pinManager.ts
```diff
@@ -0,0 +1,92 @@
...
+	 * @param {*} componentList
+	 * @param {Pin[]} pins
+	 * @returns {Pin[]}
+	 * @memberof PinManager
+	 */
+	handleNameChanges(componentList: any, pins: Pin[]): Pin[] {
+		return pins.map((pin: Pin) => {
+			if (typeof pin.component === "undefined") return pin;
+
+			let componentConfig = componentList[pin.component];
```
amaroks:
```suggestion
			const componentConfig = componentList[pin.component];
```

PR #416, opened 2019-06-25
Fix #14694 When a folder dragged over "My Apps" or "Favorites" move to the top of the list

amaroks:
Fix #id [14694]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/14694/details)

**Description of change**
* Track mouse Y position
* Move folder to the top/bottom of list based on mouseUp position

**Description of testing**

1. Check out this branch and switch to new app launcher `My Apps`
2. Start fresh and create 3-4 folders
3. Drag a folder from center or bottom all the way to the top or top outside the window
- [x] The folder becomes first in the list and below `My apps` and `Favorites`
4. Drag a folder from center or top all the way to the bottom or bottom outside the window
- [x] The folder becomes last in the folders list

`NOTE: When dragging the folder away from the window and releasing it, it will first return to where it was and then get to the top or bottom of the list, a limitation in current HTML5 drag and drop`

See screenshot for expected results.

myapps-reorder folders

### DIFF
--- a/src-built-in/components/myApps/src/components/FoldersList.jsx
+++ b/src-built-in/components/myApps/src/components/FoldersList.jsx
@@ -22,6 +22,8 @@ export default class FoldersList extends React.Component {
 			folderNameInput: '',
 			isNameError: false
 		}
+		// Reference to ontainer element of folder list
+		this.listDiv = React.createRef();
 		this.renameFolder = this.renameFolder.bind(this)
 		this.changeFolderName = this.changeFolderName.bind(this)
 		this.onFoldersListUpdate = this.onFoldersListUpdate.bind(this)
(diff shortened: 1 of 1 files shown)


amaroks:
You probably found another bug. I will look into it, however you're not testing what this change was made for :) see my gif its mostly about dragging a folder away of the window or on top of my apps and favorites. Does this reordering thing work every time on 3.10.0 for you?

zona-ciq:
> You probably found another bug. I will look into it, however you're not testing what this change was made for :) see my gif its mostly about dragging a folder away of the window or on top of my apps and favorites. Does this reordering thing work every time on 3.10.0 for you?

I talked to Kyle and the expectation is that the folder will drop wherever user releases the mouse, and if that is above or below the folders, it should go on the top or bottom of the folder list. Kyle will update the card to reflect that.

PR #435, opened 2019-07-25
Fix 15127 My Apps is vulnerable to components with invalid config

amaroks:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/15127/details)

**Description of change**
* Catch exception when thrown during access of non existing `launchableByUser`

**Description of testing**
1. Check out this branch run `npm run dev`
1. Editing toolbar/config.json line 50 set menuType to `My Apps`
1. Edit components.json remove `foreign` key from Welcome Component
1. Run the seed `npm run dev`
- [x] My Apps launcher still works
- [x] Welcome components is not there
- [x] No My Apps errors in central logger

### DIFF
--- a/src-built-in/components/myApps/src/stores/StoreActions.js
+++ b/src-built-in/components/myApps/src/stores/StoreActions.js
@@ -1,3 +1,4 @@
+import _get from 'lodash.get';
 import { getStore } from "./LauncherStore";
 import AppDirectory from "../modules/AppDirectory";
 import FDC3 from "../modules/FDC3";
@@ -175,12 +176,7 @@ function loadInstalledConfigComponents(cb = Function.prototype) {
 			// If the app is already in our list move on
 			if (appInAppList(componentName)) return;
 			const component = componentList[componentName];
-			let launchableByUser;
-			try {
-				launchableByUser = component.foreign.components["App Launcher"].launchableByUser;
-			} catch (error) {
-				launchableByUser = false
-			}
+			const launchableByUser = _get(component, 'foreign.components.App Launcher.launchableByUser');
 			// Make sure the app is launchable by user
 			if (launchableByUser) {
 				data.configComponents[componentName] = {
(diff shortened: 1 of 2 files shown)


src-built-in/components/myApps/src/stores/StoreActions.js
```diff
@@ -174,9 +174,15 @@ function loadInstalledConfigComponents(cb = Function.prototype) {
 		componentNameList.map(componentName => {
 			// If the app is already in our list move on
 			if (appInAppList(componentName)) return;
-			let component = componentList[componentName];
+			const component = componentList[componentName];
+			let launchableByUser;
+			try {
+				launchableByUser = component.foreign.components["App Launcher"].launchableByUser;
```
stephen-ciq: Instead of using `try/catch`, you can use `lodash.get`.  I think it is a little cleaner.  See an example in some code I wrote here: https://github.com/ChartIQ/finsemble/pull/1435/files#diff-b0f6bb271af0bec4b31c620511df3755R249

amaroks:
So I'm a big fan of lodash and have argued a lot about it and was one of the devs who kept asking to have it. We do have it in FEA, Finsemble but unfortunately we don't have it in seed dependencies. And that is why I actually did not use it. @carterbs Should I just install `lodash/get` ?

carterbs:
> So I'm a big fan of lodash and have argued a lot about it and was one of the devs who kept asking to have it. We do have it in FEA, Finsemble but unfortunately we don't have it in seed dependencies. And as you know, adding dependencies requires approval, that is why I actually did not use it. @carterbs Should I just install `lodash/get` ?

Just got the go-ahead from ian to use lodash.get here. Please keep its use to a single file so we can refactor it easily later. Apparently there is a new feature of javascript that will be here in the next months/years that will allow us to do the same thing as lodash.get.

src-built-in/components/myApps/src/stores/StoreActions.js
Garbee: I would concur with changing this to lodash's get currently. Ian is assessing options to address the need of this, including possibly using the [optional chaining](https://github.com/tc39/proposal-optional-chaining) that will be built into JavaScript soon. So having one easy path of things to look for to update if that move happens would be nicer than needing to go filter between lodash get usage and any possibly try/catches that exist only for this purpose.

PR #439, opened 2019-08-05
Set up close request confirmation dialog when user right clicks to close finsemble toolbar

erik-ciq:
feat: #id [12600](https://chartiq.kanbanize.com/ctrl_board/18/cards/12600/details/)

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/12600/details)

**Description of change**
* Adds a close confirmation dialog when a user right clicks to close finsemble toolbar. Previously there was no dialog, and choosing close would close the toolbar (with no way of getting it back). 

**Description of testing**
1. Checkout `planned/4.0.0` in `finsemble` and `finsemble-electron-adapter`
2. Checkout `12600-toolbar-close-confirm` in `finsemble-seed`
3. Launch finsemble
4. Right click the left hand side of the finsemble toolbar

image

5. Choose close
6. A dialog should appear. Test both cases.

-> Clicking cancel should do nothing
-> Clicking Shut down should should down finsemble

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
         "@chartiq/finsemble-react-controls": "3.7.0",
         "async": "2.6.2",
         "babel-minify-webpack-plugin": "0.3.1",
-        "lodash.get": "^4.4.2",
+        "lodash.get": "4.4.2",
         "compression": "1.7.3",
         "file-loader": "1.1.11",
         "lodash.clonedeep": "4.5.0",
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
         "@chartiq/finsemble-react-controls": "3.7.0",
         "async": "2.6.2",
         "babel-minify-webpack-plugin": "0.3.1",
+        "lodash.get": "^4.4.2",
         "compression": "1.7.3",
         "file-loader": "1.1.11",
         "lodash.clonedeep": "4.5.0",
--- a/src-built-in/components/toolbar/src/app.jsx
+++ b/src-built-in/components/toolbar/src/app.jsx
@@ -6,14 +6,21 @@
 // Static vs Dynamic Toolbar
 import Toolbar from "./dynamicToolbar";
 // import Toolbar from "./staticToolbar";
-//Band-aid. Openfin not respecting small bounds on startup.
-if (!fin.container || fin.container != "electron") {
-  fin.desktop.main(() => {
-    let finWindow = fin.desktop.Window.getCurrent();
+
+function FSBLReady() {
+  //Band-aid. Openfin not respecting small bounds on startup.
+  if (!FSBL.System.container || FSBL.System.container != "electron") {
+    let finWindow = FSBL.System.Window.getCurrent();
     finWindow.getOptions((opts) => {
       if (opts.smallWindow) {
         finWindow.setBounds(opts.defaultLeft, opts.defaultTop, opts.defaultWidth, opts.defaultHeight);
       }
     })
-  })
+  }
+}
+
+if (window.FSBL && FSBL.addEventListener) {
+  FSBL.addEventListener("onReady", FSBLReady);
+} else {
+  window.addEventListener("FSBLReady", FSBLReady);
 }
--- a/src-built-in/components/toolbar/stores/toolbarStore.js
+++ b/src-built-in/components/toolbar/stores/toolbarStore.js
@@ -227,7 +227,7 @@ class _ToolbarStore {
 			self.hideToolbar();
 		});
 
-		fin.desktop.Window.getCurrent().addEventListener("close-requested", () => this.closeRequestedHandler());
+		FSBL.System.Window.getCurrent().addEventListener("close-requested", () => this.closeRequestedHandler());
 	}
 
 	/**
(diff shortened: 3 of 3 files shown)


src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -225,6 +225,23 @@ class _ToolbarStore {
...
+	/**
+	 * handles a close request by showing the user a dialog prompting for shutdown.
+	 * affirmative will shutdown finsemble, negative will do nothing.
+	 */
+	async closeRequestedHandler() {
+		const args = await this.confirmCloseToolbar();
+		if (args.err) {
+			Logger.system.log(`Error received on confirm close: ${args.err} Continuing.`);
+		}
+		if (args.result.choice === 'affirmative') {
```
amaroks: Should it really continue and get to this line if there is `args.err`?

erik-ciq: I figure if there is both an error, and an affirmative shutdown, then the affirmative shutdown is important enough to just proceed despite the error.

amaroks: Okay I was thinking that maybe if there is an error then `args.result` will not be there.

erik-ciq: It's definitely possible, but in that case it would just log the error and do nothing.

src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -275,6 +295,23 @@ class _ToolbarStore {
...
+	 */
+	confirmCloseToolbar = () => {
+		const dialogParams = {
+			title: "Confirm Shutdown",
+			question: "Do you wish to shut down finsemble?",
+			affirmativeResponseLabel: "Shut down",
+			negativeResponseLabel: "Cancel"
+		};
+		return new Promise(resolve => {
+			FSBL.Clients.DialogManager.open("yesNo", dialogParams, (err, result) => resolve({ err, result }));
```
Garbee: Since this is a new promise being added, we should not resolve with an error. If an error happens, reject. Resolve if successful. This is how promises are *intended* to be used. We have tons of existing problems and hard to understand things because of this practice not being followed.

Starting to guard against this practice now gets us in the habit of doing future promises better initially rather than just waiting for it to be decided to do things right and causing more code-churn later.

stephen-ciq: I agree with Garbee on this one.

src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -225,6 +225,26 @@ class _ToolbarStore {
...
+	/**
+	 * handles a close request by showing the user a dialog prompting for shutdown.
+	 * affirmative will shutdown finsemble, negative will do nothing.
+	 */
+	async closeRequestedHandler() {
+		const args = await this.confirmCloseToolbar();
+		// proceed even if there is an error in case the user selected to shut down. Shut down despite error.
+		// do nothing if there was an error or they chose cancel.
+		if (args.err) {
+			Logger.system.log(`Error received on confirm close: ${args.err} Continuing.`);
```
stephen-ciq:
```suggestion
			Logger.system.log(`Error received on confirm close: ${args.err}. Continuing.`);
```

src-built-in/components/toolbar/stores/toolbarStore.js
mesner: This was already decided. Never throw and error.  Never reject a promise.  If you can convince Terry otherwise, by all means.

erik-ciq:
@amaroks @stephen-ciq  please re-review.

- [x] Fix lodash.get import
- [x] Fix two cancel buttons

src-built-in/components/toolbar/stores/toolbarStore.js
```diff
@@ -225,6 +226,26 @@ class _ToolbarStore {
 		FSBL.Clients.HotkeyClient.addGlobalHotkey(["ctrl", "alt", "h"], () => {
 			self.hideToolbar();
 		});
+
+		fin.desktop.Window.getCurrent().addEventListener("close-requested", () => this.closeRequestedHandler());
```
amaroks:
```suggestion
		fin.desktop.Window.getCurrent().addEventListener("close-requested", this.closeRequestedHandler);
```

src-built-in/components/toolbar/src/app.jsx
```diff
@@ -6,14 +6,21 @@
...
 // import Toolbar from "./staticToolbar";
-//Band-aid. Openfin not respecting small bounds on startup.
-if (!fin.container || fin.container != "electron") {
-  fin.desktop.main(() => {
-    let finWindow = fin.desktop.Window.getCurrent();
+
+function FSBLReady() {
+  //Band-aid. Openfin not respecting small bounds on startup.
+  if (!FSBL.System.container || FSBL.System.container != "electron") {
+    let finWindow = FSBL.System.Window.getCurrent();
```
stephen-ciq: I know we are removing references to `fin` in 4.0, but should we also modify names like `finWindow` or is that okay as it is?

carterbs: That's fine. I'll follow up with my rationale.

package.json
```diff
@@ -39,6 +39,7 @@
         "@chartiq/finsemble-react-controls": "3.7.0",
         "async": "2.6.2",
         "babel-minify-webpack-plugin": "0.3.1",
+        "lodash.get": "^4.4.2",
```
carterbs:
```suggestion
        "lodash.get": "4.4.2",
```

PR #440, opened 2019-08-06
15092 Swap openfin-launcher with hadouken-js-adapter

amaroks:
Upgrade from the openfin-launcher to the Hadouken-adapter to officially support 10.15.3.

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/15092/details)

**Description of change**
* Remove openfin-launcher, add hadouken-js-adapter
* Use hadouken API instead of openfin launcher

**Test 1**
1. Set container to `openfin` under `server-environment-startup.json`
1.  `npm install`
1. `npm run dev`

- [ ] Finsemble launches without problems

**Test 2**
1. `npm run nobuild:dev`
- [ ] Finsemble launches without problems

**Test 3**
1. Delete openfin folder from %appData%
1. `npm run dev`
- [ ] Takes some time and eventually installs openfin runtime
- [ ] Finsemble launches without problems


configs/openfin/manifest-local.json
```diff
@@ -58,8 +58,8 @@
     "fileName": "ChartIQ-local-installer",
     "appAssets": [
         {
-            "src": "https://files.chartiq.com/finsemble-dev/Assimilation_3_11_0.zip",
-            "version": "3.11.0_48",
+            "src": "https://assets.finsemble.com/assimilation/dev/assimilation_pr50.zip",
```
stephen-ciq: This looks like we are referencing a PR in a manifest file, which seems strange.  But I don't know the context.  Could you explain why this change is necessary?

package.json
```diff
@@ -41,6 +40,7 @@
         "babel-minify-webpack-plugin": "0.3.1",
         "compression": "1.7.3",
         "file-loader": "1.1.11",
+        "hadouken-js-adapter": "^1.43.1",
```
stephen-ciq: are you sure that the caret is necessary?  Sometimes npm adds these.  But I don't know if it's necessary here.

amaroks: its okay unless we want to lock versions which is usually what I like to do. I just wasn't told that we should lock versions, I talked about it once and somebody else took care of this.

configs/openfin/manifest-local.json
amaroks: Its because it was checked out from planned/3.11.0 instead of the newer  branches.

package.json
amaroks: @carterbs Should we lock package versions as soon as we install them?

PR #442, opened 2019-08-22
add requireHandshake property to manifest

mhmcclung:
fix: #id 15366

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/15366/details/)

**Description of change**
Enable FEA handshake in manifest. 

**Description of testing**
See FEA companion PR in ticket for description of testing.

### DIFF
--- a/configs/openfin/manifest-local.json
+++ b/configs/openfin/manifest-local.json
@@ -3,7 +3,8 @@
     "startup_app": {
         "name": "ChartIQ Local",
         "url": "http://localhost:3375/finsemble/components/system/serviceManager/serviceManager.html",
-        "uuid": "Finsemble",
+		"uuid": "Finsemble",
+		"requireHandshake": true,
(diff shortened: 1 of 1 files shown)


configs/openfin/manifest-local.json
```diff
@@ -3,7 +3,8 @@
     "startup_app": {
         "name": "ChartIQ Local",
         "url": "http://localhost:3375/finsemble/components/system/serviceManager/serviceManager.html",
-        "uuid": "Finsemble",
+		"uuid": "Finsemble",
+		"requireHandshake": true,
```
zona-ciq: Inside a comment in FEA it says "by default the check is disabled", so should this flag be false?

mhmcclung: After discussing with Brad and Ian it was decided to have the value default to false for general compatibility (including if FEA is used without Finsemble).  But we'll always want to have it enabled with Finsemble since it provides one extra check that everything comes up correctly (and puts out notification if it doesn't).   So setting to true is correct.

PR #444, opened 2019-09-03
removed all "fixed toolbar" code

mhmcclung:
fix: #id 15915

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/15915/details/)

**Description of change**
* Removed all "Fixed Toolbar" code from seed (there was nothing to remove in Finsemble)

**Description of testing**
1. Start Finsemble
1. Open Preferences
1. Confirm no toolbar related preferences are shown and no errors in central logger


d4hines:
There's a tradeoff here: add the message at the cost of more code, or just omit the message at the cost of potentially confused users. I'd lean towards less code - the purpose of this was to simplify, and I want to squeeze as many removed lines out of this PR as we can.

d4hines:
Mike and I both tested each others work btw.

d4hines:
I'm curious why the toolbar didn't come up though... That doesn't seem right. Investigating...

clausjs:
> There's a tradeoff here: add the message at the cost of more code, or just omit the message at the cost of potentially confused users. I'd lean towards less code - the purpose of this was to simplify, and I want to squeeze as many removed lines out of this PR as we can.

I certainly can agree with this explanation. Again, just wanted to make sure we discuss it. This seems more like a business decision that those in product would at least like to have an opinion on, or so I'd think. Its great to lose code when we can, but who it will affect and how harshly are more important (probably)

mhmcclung:
I was going on the assumption that no customers use the fixed-toolbar code.  But perhaps that's not true.  Either way, I can easily put in a 2-line config override to handle this case.

mhmcclung:
After more discussion here's two conclusions: 
1) we only officially support one config setting for full toolbar functionality (i.e. dockable set and spawnToAllMonitors false).  2) If that config setting isn't enabled then the "toolbar" should operate like any other component (which can also use variations of dockable and spawnToMonitor settingss).  

Currently #2 above is not true because spawnOnAllMonitors is broken (independent of the toolbar) - see https://chartiq.kanbanize.com/ctrl_board/48/cards/16027/details/.  Also, if dockable is set to false then the toolbar isn't shown on startup because autoshow is set to false; however setting autoshow to true for this case fixes the problem.  

So bottom line is I didn't change any code for this PR.  I only pushed one config change, deleting an autoshow entry in the toolbar that a duplicate and incorrectly positioned (and missleading).

clausjs:
This works, but I'm not sure how I feel about the decision. I understand minimizing code, but realistically finsemble-seed from a sales perspective is very much _the_ product. If that's the case, we need our presentation components to perform well and look nice. It doesn't matter how well Finsemble works if the UI being sold with it is off-putting. 

That being said, its not going to be inherently obvious to a user that if (for whatever reason) they wanted a fixed toolbar (since there's nothing telling them its no longer an option) they would need to set `dockable: false` and `options: autoShow: true` (when they didn't have to do this before). I _really_ think this is a decision that needs to be discussed with product, and while we may not like their answer because it might require more code it would at least lead to a smoother more user friendly experience. That's just my two cents though

mhmcclung:
@sgd2z I think the hold up on this ticket comes back to requirements, so we'll need a call on what's next.   Garbee also had some related comments in a slack conversation.

Garbee:
We need to make sure these broken config parameters are properly documented as no longer supported. So long as that is done for this release, then we aren't addressing the config problems where the hidden toolbar is a problem.

PR #445, opened 2019-09-05
15957 api testing components REVIEW ONLY

ethan-ciq:

**Description of change**
* Changes
Create 2 components for api testing (apitest1 and apitest2)

**Description of testing**
1. Step 1.
1. Step 2.
1. [x] Thing happened.
1. [ ] Thing did not happen.
1. Step 3


configs/application/components.json
```diff
@@ -185,6 +185,81 @@
...
+		},
+		"apitest1": {
+			"window": {
+				"url": "$applicationRoot/components/apitest/test1.html",
+				"frame": false,
+				"resizable": true,
+				"autoShow": true,
+				"top": "center",
+				"left": "center",
+				"width": 800,
```
NinjaManatee: Make the component wider so the tabs fit in the initial width.
```suggestion
				"width": 925,
```

src-built-in/components/apitest/test1.html
```diff
@@ -0,0 +1,77 @@
...
+		<script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
+		<link
+			href="https://fonts.googleapis.com/css?family=Open+Sans+Condensed"
+			rel="stylesheet"
+		/>
+		<script src="../../vendor.bundle.js"></script>
+	</head>
+
+	<body>
+		<label for="symbolInput" id="symbolLabel">Input text</label>
```
NinjaManatee: What is input text used for, it isn't clear.

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+	$("#Search").append(search);
+}
+/* SeachClient Functions*/
+function triggerSearch() {
+	if (document.getElementById('symbolInput').value != '')
+		searchClient.search({
+			text: document.getElementById('symbolInput').value
+		}, function (err, response) {
+			if (!err) {
+				alert('Sussceed. See console for detail.')
```
NinjaManatee: Could you add test next to the button with message rather than using alerts?
```suggestion
				alert('Success. See console for detail.')
```

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+	})
+}
+
+function triggerGetStoreValue() {
+	distributedStoreClient.getStore({
+		store: 'testDs1',
+	}, function (err, store) {
+		if (!err) {
+			store.getValue({
+				field: 'field1'
```
NinjaManatee: This should retrieve a value that actually exists.

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+			}, function (err, value) {
+				alert('Field1: ' + value)
+			})
+		} else
+			alert('No such store.')
+	})
+}
+
+function triggerRemoveStore() {
```
NinjaManatee: The order of the test may need to be rearranged. If I start at the top and work my way down, the store is added and removed before I start adding/removing values from the store.

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+/* Distributed store client functions */
+function triggerSetStoreValue(field) {
+	distributedStoreClient.getStore({
+		store: 'testDs1',
+	}, function (err, store) {
+		if (!err) {
+			store.setValue({
+				field: field,
+				value: document.getElementById('symbolInput').value
+			}, function (err) {})
```
NinjaManatee: This test has no indication of success.

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+	getComponentDefaultConfig.click(function () {
+		triggerGetComponentDefaultConfig();
+	});
+	$("#Launcher").append(getComponentDefaultConfig);
+	var getComponentList = $("<getComponentList>Get Component List</getComponentList>");
+	getComponentList.click(function () {
+		triggerGetComponentList();
+	});
+	$("#Launcher").append(getComponentList);
+	var getComponentsThatCanReceiveDataTypes = $("<getComponentsThatCanReceiveDataTypes>Get Components That Can Receive Data Types</getComponentsThatCanReceiveDataTypes>");
```
NinjaManatee: The button isn't wide enough for this text.

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+			'rsrchx.report': {
+				symbol: event.target.value
+			},
+			'symbol': event.target.value
+		};
+		dragdropClient.dragStartWithData(event, data);
+	})
+
+	routerClient.onReady(function () {
+		console.log('RouterClient is ready')
```
NinjaManatee: In general, I think it would be better to have the output available from the UI rather than having to open the dev console.

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+		},
+		function (err, response) {
+			if (!err) {
+				test2SpawnResopnse = response
+			} else
+				alert('Spawn error')
+		})
+}
+
+function triggerShowWindow() {
```
NinjaManatee: I don't get a result for this test. Not sure what happened.

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+function triggerOpenLinkerWindow() {
+	linkerClient.openLinkerWindow(function (err, response) {
+		if (!err) {
+			console.log(response)
+		}
+	})
+}
+
+function triggerStartOnStateChange() {
+	linkerClient.onStateChange(function (err, response) {
```
NinjaManatee: This doesn't give a success message.

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+		data: document.getElementById('symbolInput').value
+	}, function (err) {
+		if (!err) {
+			//alert('Publish sussceed.')
+		}
+	})
+}
+
+function triggerOpenLinkerWindow() {
+	linkerClient.openLinkerWindow(function (err, response) {
```
NinjaManatee: This doesn't appear to work.

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+}
+
+function triggerPublish() {
+	routerClient.publish('symbol', {
+		'symbol': document.getElementById('symbolInput').value
+	})
+}
+
+function triggerRemovePubSubResponder() {
+	routerClient.removePubSubResponder('symbol')
```
NinjaManatee: No success or failure for this.

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+	}, function (error, queryResponseMessage) {
+		if (!error) {
+			// process income query response message
+			alert("Router client query respond: " + queryResponseMessage.data)
+		}
+	});
+}
+
+function triggerPublish() {
+	routerClient.publish('symbol', {
```
NinjaManatee: No success or failure for this

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+}
+
+function triggerTransmit() {
+	routerClient.transmit('symbol', {
+		'data': document.getElementById('symbolInput').value
+	})
+}
+
+function triggerQuery() {
+	routerClient.query("symbol", {
```
NinjaManatee: No success message

src-built-in/components/apitest/test1.js
```diff
@@ -0,0 +1,1538 @@
...
+}
+
+/* Router client functions */
+function triggerDisconnectAll() {
+	routerClient.disconnectAll()
+	alert('Disconnected all.')
+}
+
+function triggerTransmit() {
+	routerClient.transmit('symbol', {
```
NinjaManatee: No success or failure message.

src-built-in/components/apitest/test1.html
ethan-ciq: My intention is to allow the tester to input some data to test some data related apis. E.g. pub sub...

src-built-in/components/apitest/test1.js
ethan-ciq: The createStore api should be invoked before this. Should there be some instructions to let tester know the sequence of the buttons?

src-built-in/components/apitest/test1.html
NinjaManatee: I discovered some of this when going through the tests. You may want to put a comment on the form or a placeholder in the text field.

src-built-in/components/apitest/test1.js
NinjaManatee: I was working down the list of buttons, so if you changed the order so that the value was added before this button, then I think it would be OK.

ethan-ciq:
**JSON.stringify(data, null, "\t");**
- Done

**For the "Get Components that can Receive data types" test, you should make the data type the one received by API Testing 2 so that a value is returned. Symbol will work in the sales demo, but that data type will work where ever the testing config is loaded.**
- Added config. But is this config(advertiseReceivers) deprecated?

**For the linker, you should move "Unlink group 1" to the bottom so "Linker Publish" sends data to "API Testing 2".** 
- Done

**Does "Open Linker Window" work for you?** 
- Not working for me too. I dont know what I am going to expect from this API since there is limited info in the API doc on this API

**does "Start on state change" print anything when there is a state change? I don't see anything happen when I link/unlink to the channel API TESTING 1 is on.**
- I found that it wont work for the first state change. If you try once more link/unlink. The stateChanged callback will be called. Is this a bug?

**I don't see any messages in the Central Logger for the logging tests. I saw the error in the dev console, though. Not sure why this isn't working.**
- I think it is because there are too many log messages. Added unique wording for bettering isolation of the messages.

**I think the router tests need some work. They don't seem to exercise the router.**
- Added proper instruction to guide the sequence of testing

**For the Storage tests, some say to look in indexedDB, but I'm not sure what I should be looking for.**
- Add proper instruction to see the result

**For Windows and UI, it doesn't seem like Bring to Front works.** 
- It seems this API is difficult to be tested. When you click on the window it should be bring to front automatically

**Get component state doesn't seem to get the state set above. 
"Get Window Groups" doesn't appear to do anything.
In Workspace, "Auto Arrange", "Get active workspace", "Import workspace", "Minimize All", "Remove workspace", "Save workspace", "Save workspace as" and "Switch to Workspace" don't appear to do anything.**
- These functions are working. But seems they are not working after "Set User".

**The drag and drop data doesnt' appear to get dropped in API Testing 2.** 
- you can drag from the textbox or "Drag to share" in the toolbar of API Test 1 to API Test 2

**"Add store listeners" says no such store, but clicking "Set Store" doesn't resolve. Am I missing something?**
- You have to create store first in API Test 1. Added proper message.

Will be moved to SD after meeting with Joel.

NinjaManatee:
For the "Get Components that can Receive data types" test, you should make the data type the one received by API Testing 2 so that a value is returned. Symbol will work in the sales demo, but that data type will work where ever the testing config is loaded.

- I didn't look at how you implemented this, only at the end result. If you think you were advertising the receivers correctly, then we should take another look to make sure that a) you were using the API correctly and b) that the documentation led you to the correct usage and c) the API works as expected.

Does "Open Linker Window" work for you?
- This didn't work for me. I've seen this working before. Lets's look into why this isn't working. 

I found that it wont work for the first state change. If you try once more link/unlink. The stateChanged callback will be called. Is this a bug?
- This sounds like a bug, but it is worth further investigation. 

It seems this API is difficult to be tested. When you click on the window it should be bring to front automatically.
- There may be an issue with the API or the documentation here. We should look into it more.

These functions are working. But seems they are not working after "Set User".
- If I remember correctly, calling `setUser` after login causes issues, and I think there is a ticket for it to throw an error after login. Let's take that out for now.

I will test again for the other issues, but I think the investigate items above should have tickets created in Kanbanize for us to look into.

ethan-ciq:
"openLinkerWindow" should has a bug.
Its calling this.routerClient.query("Finsemble.LinkerWindow.Show"..... and it returned the following. 
"No responder for Finsemble.LinkerWindow.Show"

ethan-ciq:
These components moved to sales-demo

PR #450, opened 2019-09-13
fix: #id 15143 Error returned is object rather than string.

gerard-ciq:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/21/cards/15143/details/)

**Description of change**
Testing error message done on object rather than string or array

**Description of testing**

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -37,7 +37,7 @@ const { launch, connect } = require('hadouken-js-adapter');
 		logToTerminal(`kill: running: ${command}...`);
 
 		exec(command, err => {
-			if (err && !err.includes(error)) {
+			if (err && err.message && !err.message.includes(error)) {
 				console.error(errorOutColor(err));
 			}
 


NinjaManatee:
@gerard-ciq Could you update the testing procedures?

gulpfile.js
```diff
@@ -37,7 +37,7 @@ const { launch, connect } = require('hadouken-js-adapter');
 		logToTerminal(`kill: running: ${command}...`);
 
 		exec(command, err => {
-			if (err && !err.includes(error)) {
+			if (err && err.message && !err.message.includes(error)) {
```
NinjaManatee: Could there be a case where err is a string instead of an Error object?

gerard-ciq: I don't think so. Unless I am reading wrong (which does happen!) it does not look like it from the docs.

https://nodejs.org/docs/latest-v10.x/api/child_process.html#child_process_child_process_exec_command_options_callback

I checked version 8 too
https://nodejs.org/docs/v0.8.28/api/child_process.html#child_process_child_process_exec_command_options_callback

gerard-ciq:
> @gerard-ciq Could you update the testing procedures?

hahaha.. I was hoping nobody would ask this. I couldn't get an error to occur because I could not get that section to execute (somehow natwest did though). So I had to move the execution somewhere else.

PR #458, opened 2019-09-30
13844 dont use deprecated config

kriswest:
chore: #id [13844]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/21/cards/13844/details/)

**Description of change**
* Remove deprecated config flags from example components and presentation components
* Replace with current flags.
* Enables Documentation and Finsemble tutorial components in launcher menu - seems reasonable for them to be launchable if included in config.

**Description of testing**
1. Build and run finsemble
1. Run through normal tests for affected components (all examples + Toolbar + all UI panels and dialogs)
1. Establish there is no unexpected or abnormal behavior (that you don't get on the base seed project without these changes)


configs/application/components.json
```diff
@@ -14,26 +14,32 @@
...
-					"dockingService": {
-						"canGroup": true,
-						"isArrangable": true
+					"windowService": {
+						"allowSnapping": true,
+						"allowGrouping": true,
+						"allowTabbing": true,
+						"allowTiling": true,
+						"allowAutoArrange": true,
+						"allowMinimize": true
```
gerard-ciq: Question. Are these all false by default or are you just putting them in here for illustration?

configs/application/components.json
```diff
@@ -14,26 +14,32 @@
 				"height": 432
 			},
 			"component": {
-				"inject": false,
+				"displayName": "Welcome Component",
 				"spawnOnStartup": false,
+				"inject": false,
```
NinjaManatee: Is inject false needed? I feel like we should remove it since inject isn't the preferred mechanism.

configs/application/components.json
```diff
@@ -106,18 +108,28 @@
...
+						"allowMinimize": true
+					},
+					"workspaceService": {
+						"persistURL": true
 					}
 				},
 				"components": {
 					"App Launcher": {
-						"launchableByUser": false
+						"launchableByUser": true
```
NinjaManatee: Have you asked whether it is OK to put this component in the menu? It used to be available but was removed, so I am reluctant to re-add it with out confirmation.

configs/application/components.json
```diff
@@ -133,18 +145,28 @@
...
```
NinjaManatee: Same as above

configs/application/components.json
kriswest: These are true by default and yes included for illustration - this is the config most often copy & pasted so it makes sense to surface some common options (IMHO).

kriswest:
@NinjaManatee dockingService references in windowTitleBarStore.js are backwards compatibility code (windowService preferred). However, router (query responder) topics like `DockingService.maximizeWindow` remain current (they are setup in DockingMain).

ignoreTilingAndTabbingRequests  references are the same (allowTabbing preferred in that case).

configs/application/components.json
kriswest: Agreed, removed.

configs/application/components.json
kriswest: I plan to ask when creating the review & merge ticket for this PR, I have a very vague memory that they were removed due to issues with the titlebar long ago (probably around the time I joined). I added them in to test them - but once done they seemed a reasonable addition.

zona-ciq: Those should be set to false. We don't want to launch "getting started tutorial" or "finsemble documentation" from the app menu. I'll push up the change.

configs/application/components.json
```diff
@@ -106,18 +107,28 @@
 			},
 			"foreign": {
 				"services": {
-					"dockingService": {
-						"isArrangable": true
+					"windowService": {
```
zona-ciq: We don't need to set windowService for this component.

configs/application/components.json
```diff
@@ -133,18 +144,28 @@
 			"component": {},
 			"foreign": {
 				"services": {
-					"dockingService": {
-						"isArrangable": true
+					"windowService": {
```
zona-ciq: Same as above. This can be taken out

configs/application/presentationComponents.json
```diff
@@ -546,16 +545,18 @@
...
-				"canMaximize": false
+				"spawnOnAllMonitors": false
 			},
 			"foreign": {
 				"services": {
-					"dockingService": {
-						"isArrangable": false,
-						"ignoreSnappingRequests": true,
-						"ignoreTilingAndTabbingRequests": true
+					"windowService": {
```
zona-ciq: These can be deleted as well since they are all default options. You showed a sample those options in the welcome component and users can also find the available options in our documentation, so it's probably not necessary to put them here again.

amaroks:
@kriswest You will need to update the base branch. I don't think you intentionally want to merge into master :) Once you do it will be easier to review changes.

kriswest:
@amaroks I don't actually know (when implementing a Review and Merge ticket) which base to point it at, which is why it will usually be pointed at master. Once you guys pick it up and plan it for a release,  then the appropriate base is known. My assumption has been that you lot will change the base yourselves once the correct one has been identified ;-)

Have changed to planned/3.14

kriswest:
THough if there is time it'd be great for this to be in 3.13 @BradleyCarter @sgd2z

carterbs:
> All looks good except for a couple of config settings (see inline comment). Thanks for making this update change! I'll push up the changes since they're pretty minor.
> 
> @BradleyCarter This card is to be merged in to Master. Should it be attached to a release?

3.14.0 is good

zona-ciq:
After discussion with Brad, we changed the release branch this PR will go in to planned/4.0.0

PR #465, opened 2019-10-02
[16445] Adding oauthResponse.html back

sonyl-ciq:
**fix: #id 16445**

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/21/cards/16445/details/)

**Description of change**

> The OAuth documentation references the oauthResponse.html file, but it doesn't exist in the seed. File does exist in the sales demo.

* Changes
  * Duplicated `finsemble-sales-demo/src/components/authentication/oauthResponse.html` to `finsemble-seed/src-built-in/components/authentication/oauthResponse.html`

**NOTE**: as using this requires a great deal of application-specific logic (what oAuth endpoints, what oAuth service, etc.) so testing will be involved. This heavily depends upon the client's implementation, so a skeleton of the process is provided below.

**Description of testing**

1. I followed [the authentication tutorial](https://documentation.chartiq.com/finsemble/tutorial-Authentication.html#oauth-2-0-openid). 
    1. For the purposes of testing, I included [googleapis](https://www.npmjs.com/package/googleapis) to handle the oAuth work for the server-side. If using Google, run `npm install googleapis`. Version tested was `^43.0.0`. This is implementation-specific and thus not included in `package.json`.
    1. `server/server.js` can be modified to accept an incoming `/authenticate` route for the `backchannel_endpoint`. As this contains secret keys, this must be server side. I have a secret gist for this with `server/server.js` and  `manifest-local.json` containing this code. Again, this is implementation-specific so code is not included in the PR.
1. Start Finsemble
    - [ ] oAuth window displays
    - [ ] Completing the auth flow displays:
       Screen Shot 2019-10-02 at 3 26 36 PM This displays the contents provided in `oauthResponse.html`

1. `x` out of the window
    - [ ] Electron quits with code `0`

### DIFF
--- a/src-built-in/components/authentication/oauthResponse.html
+++ b/src-built-in/components/authentication/oauthResponse.html
@@ -132,9 +132,5 @@
 		</div>
 	</div>
 	<div id="error"></div>
-
-	<script src="https://apis.google.com/js/api.js"></script>
-
-
 </body>
 </html>
--- a/src-built-in/components/authentication/oauthResponse.html
+++ b/src-built-in/components/authentication/oauthResponse.html
@@ -16,8 +16,6 @@
 	<meta name="description" content="">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
 	<script>
-		var urlParams = new URLSearchParams(window.location.search);
-		console.log('oauthresponse', urlParams)
 		/** Class for handling oAuth returns */
 		class HandleOauth {
 			constructor() {
@@ -68,10 +66,8 @@
 		const OauthClient = new HandleOauth();
 
 		FSBL.Clients.AuthenticationClient.completeOAUTH(null, null, (err, response) => {
-			
-			console.log('AuthenticationClient.completeOAUTH', response)
 			if (err) {
-				OauthClient.showError('x');
+				OauthClient.showError(err);
 			} else {
 				const credentials = response;
 				OauthClient.runBusinessLogic(credentials, () => {
@@ -131,14 +127,12 @@
 	</style>
 	<div class="fsbl-auth-top">
 		<img class="fsbl-logo" src="../../assets/img/ciq-banner-black-100x25.png" width="100" height="25">
-		<div class="fsbl-close" id="FSBL-close" onclick="closeThisWindow()">
+		<div class="fsbl-close" id="FSBL-close" onclick="OauthClient.closeThisWindow()">
 			<i class="ff-close" style=color:#666></i>
 		</div>
 	</div>
 	<div id="error"></div>
-	<script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
 
-	<script src="authentication.js"></script>
 	<script src="https://apis.google.com/js/api.js"></script>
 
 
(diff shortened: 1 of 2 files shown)


src-built-in/components/authentication/oauthResponse.html
```diff
@@ -0,0 +1,108 @@
...
+		function closeThisWindow(){
+			// When we close this window, if the user isn't logged in then it must mean we
+			// were using OAUTH to log in, and it failed, so shut down the entire app. Otherwise
+			// we were just using it to authenticate a client, in which case we just want to close
+			// this window but keep Finsemble running.
+			FSBL.Clients.AuthenticationClient.getCurrentCredentials(function(err, credentials){
+				if(!credentials){
+					FSBL.shutdownApplication();
+				}else{
+					fin.desktop.Window.getCurrent().close(true);
```
NinjaManatee: Would you replace this with a FSBL call? I think FSBL.Clients.WindowClient.getCurrent().close(true) should do it.

sonyl-ciq:
@NinjaManatee OK, will clean things up. It crossed my mind to do so, so thanks for the push. 

I haven't tested using SalesForce, but I did use Google oAuth. I'll also test with SalesForce.

src-built-in/components/authentication/oauthResponse.html
sonyl-ciq: @NinjaManatee I reworked this a bit but with regards to SalesForce, the page for that in the sales demo is very different: https://github.com/ChartIQ/finsemble-sales-demo/blob/master/src/components/salesforce/oauth2/authed.html

I might be missing something.

src-built-in/components/authentication/oauthResponse.html
```diff
@@ -0,0 +1,136 @@
...
+    * All rights reserved.
+	*/
+	
+	OAUTH Authentication example. This example uses Google as a OpenID Provider (OP).
+-->
+
+<head>
+	<meta charset="utf-8">
+	<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+	<title>Oauth Authentification</title>
```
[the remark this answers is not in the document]
sonyl-ciq: I'll fix; we should make a note to fix it in the sales demo as well.

Garbee:
I discussed this change request with Terry. I have concerns over us expanding the seed content even further with things for client developers to modify and then have conflicts during upgrades over.

Terry has agreed, that in this circumstance at least, it is fine that we not include this in the seed and instead focus on fixing the documentation. The documentation should reflect what this file should look like/do and not reference a file existing in the seed. Client developers should get the file contents as a starter point and add it in an area we don't modify to try and reduce any merge conflict problems down the line as we are doing more work in the seed structure.

NinjaManatee:
@Garbee That sounds reasonable to me.

sonyl-ciq:
@NinjaManatee would it make sense for me to take a stab at writing the doc?

NinjaManatee:
@sonyl-ciq Yes, write the initial pass for the tutorial, or extension of the current tutorial. That will make Josh's life much easier as he will just need to massage existing content. Thanks!

thorsent:
@NinjaManatee 
I've manually merged this into upcoming 5.0 branches. FYI, in case you'd rather just close this and not merge into 4.x branches

thorsent:
Old

PR #469, opened 2019-10-11
[11790] Native component freeze/breaks when tabbing on self

amaroks:
fix: #id [11790]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/11790/details)

**Description of change**
* Add check and cancel tabbing when necessary

**Description of testing**
This issue is happening when there is a single tab only. When you drag the component title and drop it on itself the component will freeze, you won't be able to move it around anymore and you won't be able to resize it. See gif below

working-selftab-freeze

To test the fix
1. Please checkout planned/3.13.0 and experience the bug
1. Check out this branch and run finsemble `npm run dev`
1. Launch a Notepad
1. Expand floating titlebar and drag the title and drop it on itself
1. Collapse floating titlebar and Try moving notepad/resizing it

- [ ] Notepad can still be tabbed with other components
- [ ] Notepad can be moved around
- [ ] Notepad can be resized

working-selftab11

### DIFF
--- a/src-built-in/components/floatingTitlebar/src/components/TabRegion.jsx
+++ b/src-built-in/components/floatingTitlebar/src/components/TabRegion.jsx
@@ -1,9 +1,7 @@
 
 
 import React from "react";
-import ReactDOM from "react-dom";
 import Tab from "./tab";
-import { FinsembleDnDContext, FinsembleDroppable } from '@chartiq/finsemble-react-controls';
 import { FinsembleHoverDetector } from "@chartiq/finsemble-react-controls";
 import Logo from "./logo";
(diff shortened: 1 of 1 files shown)


amaroks:
@meredith-ciq So this error is there in planned/3.13.0 prior the fix. It is unrelated to the freezing issue. usually when you move the window if I'm not wrong.

meredith-ciq:
> @meredith-ciq So this error is there in planned/3.13.0 prior the fix. It is unrelated to the freezing issue. usually when you move the window if I'm not wrong.

You're correct and this error is already attached to another card so no need to worry about it.

PR #474, opened 2019-10-16
fix: floating titlebar maximize error

Garbee:
When the event handler was removed in 45771b5498bf861e7388191499102faed73b83ce the binding of it was not removed.

This removes the binding, so an undefined handler is no longer bound.

fix: #16759

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/16759/details)

**Description of testing**

Maximize a native window on OpenFin that has a floating titlebar (like Notepad.) Observe no errors in the Central Logger from that action.


Garbee:
Hmm. So stacks do work differently too. I'll take a look this evening.

Garbee:
So, the drag out of a stack is a separate distinct error. It seems the same, but the source is from somewhere else in the code. Probably another event listener like this one.

Just asked Brent about this, we won't be going after the drag out in this PR. That should get its own card with it's own distinct effort.

mhmcclung:
Based on above comments I'm approving (the main problem was fixed).

PR #481, opened 2019-10-23
fix: #id 17780 App URL Validation

sonyl-ciq:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/48/cards/17780/details/)

**Description of change**
* Relaxes rules in order to allow:
    1. Custom protocols (_e.g._ `finsemble://`)
    1. Allows for port-based URLs (_e.g._ `http://localhost:3000`)

 * Revised comment

**Description of testing**

1. Enable My Apps: edit `/src-built-in/components/Toolbar/config.json` as such:
    
```js
    {
        "align": "left",
        "label": "Apps",
        "id": "app-launcher",
        "comment": "Change menuType to 'My Apps' for a preview of the finsemble app catalog!",
        "menuType": "My Apps"
    },
```
1. You may also need to add a 'foreign' element to the config of nonConfiguredComponent (as this being missing breaks My Apps...)
1. Open Apps menu item
1. Hit 'Add new app' in the bottom left
1. Enter the URL `http://localhost:3000` and submit the form
    1. [ ] `http://localhost:3000` is allowed.
    1. [ ] `fsbl://localhost` is allowed
    1. [ ] `localhost` is not allowed.

Regex test suite: https://regex101.com/r/Z85kF8/13/tests

### DIFF
--- a/src-built-in/components/myApps/src/components/AddNewAppForm.jsx
+++ b/src-built-in/components/myApps/src/components/AddNewAppForm.jsx
@@ -147,7 +147,7 @@ export default class AddNewAppForm extends React.Component {
 	 * @param {keyboardEvent} e
 	 */
 	validateURL(url) {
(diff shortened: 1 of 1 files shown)


Garbee:
FYI #487 fixed this in a more robust way than regex. So this change is most likely not needed anymore. If we are trying to get this into 3.13 then that patch should be pulled back rather than fixing regex.

PR #486, opened 2019-10-31
feat: #id 17132 validate max node version

meredith-ciq:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/17132/details)

**Description of change**
* Added validation for max node version being 10.15.3 (can be changed in the gulpfile). If validation fails an error will be logged but finsemble will still start.
* Did not use npm engines because it has a bug where the version warning doesn't log prior to npm version 6.12.0

**Description of testing**
To test without changing npm versions:
1. Note your node version, if <= 10.15.3, start finsemble and verify there are no errors in the terminal above the launching Finsemble line.
2. In the finsemble-seed gulpfile.js change MAX_NODE_VERSION to a version below your current node version
3. Launch finsemble, verify finsemble launches but there is an error in the terminal about the node version not supported

If you'd prefer you can update node to be above the supported version, though updating node is not trivial. Note that due to a bug in node 10.16 and up finsemble may or may not start, but the error should still be logged about the node version not being supported
To update Node using nvm:
1. run 'nvm install 10.16.0'
1. Make sure no finsemble build processes are running (fea, finsemble etc). In the finsemble repo, navigate to node_modules and delete the node-sass folder
1. in %APPDATA%/npm-cache delete the node-sass folder
1. npm install in the finsemble directory (make sure node-sass reinstalls, if not npm install node-sass)
1. in the fea and finsemble folders run 'npm link' (you need to set up symlinks from these folders to the new version folder in the nvm directory, seed's symlinks are connected to this folder).
1. In the finsemble-seed directory run npm link @chartiq/finsemble (and @ chartiq/finsemble-electron-adapter) to transfer seeds links to the links in your 10.16.0 nvm directory from your old node directory ( Actually this step is probably not necessary because seed seems to just link to the symlinked nvm version in Program Files, but it doesn't hurt to run)
1. Build finsemble and fea then launch finsemble. There should be an error about node 10.16.0 not being supported
1. After testing use the above steps to go back to a version of node at or before 10.15.3 (do not stay on node 10.16.0)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -60,7 +60,7 @@ const { launch, connect } = require('hadouken-js-adapter');
 		let tempVersionArray;
 		let semverObject;
 		if (typeof version !== 'string') {
-			console.log(`Version must be type string`);
+			console.log(`Version must be type string but is ${typeof version}`);
 			return;
 		}
 		// Split the version into a temp array.
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -53,46 +53,79 @@ const { launch, connect } = require('hadouken-js-adapter');
 	}
 
 	/** 
+	* Splits a string version with semantic versioning into an object with major, minor and patch versions
+	* Valid inputs are 'X.X.X' or 'vX.X.X'
+	*/
+	const createSemverObject = (version) => {
+		let tempVersionArray;
+		let semverObject;
+		if (!version || typeof version !== 'string') {
+			console.log(`Version must be type string`);
+			return;
+		}
+		// Split the version into a temp array.
+		if (version.startsWith('v')) {
+			tempVersionArray = version.split('v');
+			tempVersionArray = tempVersionArray[1].split('.');
+		} else {
+			tempVersionArray = version.split('.')
+		}
+		if (tempVersionArray)
+
+		// Convert each array element to a number and store in the object.
+		semverObject = {
+			majorVersion: Number(tempVersionArray[0]) || null,
+			minorVersion: Number(tempVersionArray[1]) || null,
+			patchVersion: Number(tempVersionArray[2]) || null,
+		}
+		// If major, minor or patch versions are missing or not a number return nothing
+		if (!semverObject.majorVersion || !semverObject.minorVersion || !semverObject.patchVersion) {
+			return;
+		}
+		return semverObject
+	}
+
+	/** 
+	* Compares two node version objects
+	* Each object is expected to contain majorVersion, minorVersion, patchVersion
+	*/
+	const compareNodeVersions = (a, b) => {
+		if (a.majorVersion !== b.majorVersion) {
+			return a.majorVersion > b.majorVersion ? 1: -1
+		}
+		if (a.minorVersion !== b.minorVersion) {
+			return a.minorVersion > b.minorVersion ? 1: -1
+		}
+		if (a.patchVersion !== b.patchVersion) {
+			return a.patchVersion > b.patchVersion ? 1: -1
+		}
+		return 0;
+	}
+
+	/** 
 	* Validates the current node version against supported node versions specified in this file
 	* Returns boolean indicating whether current node version is valid
-	* Currently only validates against a max node version
+	* Currently only validates against a max node version which must be in the format 'X.X.X' or 'vX.X.X'
 	*
 	* Note: This method is being used instead of npm engines because of an npm bug where warnings don't print
 	* This bug was resolved in npm 6.12.0 but as that is a very new version of npm and is not linked to node 10.15.3
 	* in nvm we can't assume our users have access to this version.
 	*/
 	const isNodeVersionValid = () => {
-		let ret = true;
-		// split the current node version into major, minor and patch numbers
-		let version = process.version.split('v');
-		version = version[1].split('.');		
-		const majorCurrent = Number(version[0]);
-		const minorCurrent = Number(version[1]);
-		const patchCurrent = Number(version[2]);
-
-		// split the max node version into major, minor and patch numbers
-		let maxNodeArray = MAX_NODE_VERSION.split('.');
-		if (maxNodeArray.length !== 3) {
-			console.log("Format of maximum node version must be: 'X.X.X', unable to validate node version");
-			return ret;
+		// Split the current node version into an object with major, minor and patch numbers for easier comparison.
+		// If any of these values are missing, nothing will be returned
+		let currentVersionObject = createSemverObject(process.version);
+		let maxVersionObject = createSemverObject(MAX_NODE_VERSION);
+		
+		// Only allow the check both objects exist and contain major, minor and patch versions. 
+		if (!currentVersionObject || !maxVersionObject) {
+			logToTerminal("Format of node version must be: 'X.X.X', unable to validate node version", "yellow");
+			return true;
 		}
-		const majorMAX = Number(maxNodeArray[0]);
-		const minorMAX = Number(maxNodeArray[1]);
-		const patchMAX = Number(maxNodeArray[2]);
 
-		// major version is too high
-		if (majorCurrent > majorMAX ) {
-			ret = false;
-		} else if (majorCurrent === majorMAX) {
-			// major version matches the max allowed version, minor version is too high
-			if (minorCurrent > minorMAX) {
-				ret = false;
-			// major and minor version matches the max allowed version, patch version is too high
-			} else if (minorCurrent === minorMAX && patchCurrent > patchMAX) {
-				ret = false;
-			}
-		}
-		return ret;
+		// Check if the node version is higher than the maximum allowed node version.
+		if (compareNodeVersions(currentVersionObject, maxVersionObject) == 1) return false;
+		return true;
 	}
 
 	let angularComponents;
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -70,19 +70,20 @@ const { launch, connect } = require('hadouken-js-adapter');
 		} else {
 			tempVersionArray = version.split('.')
 		}
-		if (tempVersionArray)
+		if (tempVersionArray.length === 3) {
 
-		// Convert each array element to a number and store in the object.
-		semverObject = {
-			majorVersion: Number(tempVersionArray[0]) || null,
-			minorVersion: Number(tempVersionArray[1]) || null,
-			patchVersion: Number(tempVersionArray[2]) || null,
+			// Convert each array element to a number and store in the object.
+			semverObject = {
+				majorVersion: Number(tempVersionArray[0]) || null,
+				minorVersion: Number(tempVersionArray[1]) || null,
+				patchVersion: Number(tempVersionArray[2]) || null,
+			}
+			// If major, minor or patch versions are missing or not a number return nothing
+			if (!semverObject.majorVersion || !semverObject.minorVersion || !semverObject.patchVersion) {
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -51,6 +52,49 @@ const { launch, connect } = require('hadouken-js-adapter');
...
+		version = version[1].split('.');		
+		const majorCurrent = Number(version[0]);
+		const minorCurrent = Number(version[1]);
+		const patchCurrent = Number(version[2]);
+
+		// split the max node version into major, minor and patch numbers
+		let maxNodeArray = MAX_NODE_VERSION.split('.');
+		if (maxNodeArray.length !== 3) {
+			console.log("Format of maximum node version must be: 'X.X.X', unable to validate node version");
+			return ret;
```
stephen-ciq: It looks to me like this will return `true` if we are unable to validate the node version; is that what we want?

gulpfile.js
```diff
@@ -51,6 +52,49 @@ const { launch, connect } = require('hadouken-js-adapter');
 		console.log(`[${new Date().toLocaleTimeString()}] ${chalk[color][bgcolor](msg)}.`);
 	}
 
+	/** 
+	* Validates the current node version against supported node versions specified in this file
+	* Returns true of false if current node version is valid
```
stephen-ciq:
```suggestion
	* Returns boolean indicating whether current node version is valid
```

Or you could do `Returns boolean if and only if the current node version is valid`

Or maybe `Returns true if the current node is valid and false otherwise`

gulpfile.js
stephen-ciq: Also, maybe we could make `const MAX_NODE_VERSION_ARRAY = [10,15,3]` so that we don't have to parse the constant string?

gulpfile.js
```diff
@@ -51,6 +52,49 @@ const { launch, connect } = require('hadouken-js-adapter');
...
+	* Currently only validates against a max node version
+	*
+	* Note: This method is being used instead of npm engines because of an npm bug where warnings don't print
+	* This bug was resolved in npm 6.12.0 but as that is a very new version of npm and is not linked to node 10.15.3
+	* in nvm we can't assume our users have access to this version.
+	*/
+	const isNodeVersionValid = () => {
+		let ret = true;
+		// split the current node version into major, minor and patch numbers
+		let version = process.version.split('v');
```
stephen-ciq: It seems counterintuitive to me that we are validating the `MAX_NODE_VERSION` below, which is a constant string, but not validating `process.version`.  OTOH maybe Node guarantees that it will be of a certain format?  If I were you, I might validate this string by checking that it matches the regex `v([0-9])+\.([0-9])+\.([0-9])+`.  And then extract the three groups in parentheses.

gulpfile.js
```diff
@@ -51,6 +52,49 @@ const { launch, connect } = require('hadouken-js-adapter');
...
+		if (maxNodeArray.length !== 3) {
+			console.log("Format of maximum node version must be: 'X.X.X', unable to validate node version");
+			return ret;
+		}
+		const majorMAX = Number(maxNodeArray[0]);
+		const minorMAX = Number(maxNodeArray[1]);
+		const patchMAX = Number(maxNodeArray[2]);
+
+		// major version is too high
+		if (majorCurrent > majorMAX ) {
```
stephen-ciq: So comparing "numbers" with more than one decimal point is weird.  Because there is possibly an infinite number of digits between each decimal point.

But if I were you, and I realize this contradicts the other suggestion I had, I might make it conceptually cleaner (at least to me) by making a float out of the major and minor versions, i.e.

`const max_node_version : float = 10.15;`
`const max_patch_version : integer = 3;`
`current_node_version = float(version[0] + "."  + version[1]);`
`current_patch_version = version[2];`
and then do

```
if (current_node_version > max_node_version)
  return false;
else if (current_node_version > max_node_version)
  return true;
else
  return current_patch_version <= max_patch_version;
```

gulpfile.js
meredith-ciq: It will return true if we don't define a max node version correctly in this file, which would just mean someone updated it incorrectly or removed it not that the node version we're using is bad. I figured for a configuration error we shouldn't report that the node version was wrong. If we returned false here it would log too errors which may be confusing. We can discuss a better way to handle this, ideally we could trust whoever edits this file to not mess up this value, but I wanted some type of protection.

If we started with an array it would remove one step parsing the max node version, but I figured the extra step was worth it to make the max node version more user readable.

gulpfile.js
meredith-ciq: Hmm I'd figured process.version would always be consistent. I'll have to look into that. I'm only validating the max version because we/dev users can change it.

gulpfile.js
```diff
@@ -51,6 +52,49 @@ const { launch, connect } = require('hadouken-js-adapter');
...
+		let ret = true;
+		// split the current node version into major, minor and patch numbers
+		let version = process.version.split('v');
+		version = version[1].split('.');		
+		const majorCurrent = Number(version[0]);
+		const minorCurrent = Number(version[1]);
+		const patchCurrent = Number(version[2]);
+
+		// split the max node version into major, minor and patch numbers
+		let maxNodeArray = MAX_NODE_VERSION.split('.');
```
amaroks: Why don't you just do 
```javascript
const MAX_NODE_VERSION = [10,15,3];
```
Instead of having to process the version string at a later stage? It would also be more readable this way
```javascript
const MAX_NODE_VERSION = {
   major: 10,
   minor: 15,
   patch: 3
}
```
or at least have a function that takes a string and returns parsed value, to use it for both current/max. Thoughts?

gulpfile.js
```diff
@@ -51,6 +52,49 @@ const { launch, connect } = require('hadouken-js-adapter');
...
+
+		// major version is too high
+		if (majorCurrent > majorMAX ) {
+			ret = false;
+		} else if (majorCurrent === majorMAX) {
+			// major version matches the max allowed version, minor version is too high
+			if (minorCurrent > minorMAX) {
+				ret = false;
+			// major and minor version matches the max allowed version, patch version is too high
+			} else if (minorCurrent === minorMAX && patchCurrent > patchMAX) {
```
amaroks: You're returning false  on all. major, minor, patch. If that is intentional then you don't need that many if /else statements, it could be simple done with
```javascript
const max = MAX_NODE_VERSION.replace(/\D/, '')
const current = process.version.replace(/\D/, '')
if (current > max ) {
  return false
}
```
Of maybe I'm missing something?

stephen-ciq: I think that would not not work as intended comparing
"v2.13.0"
to
"v2.2.0"

amaroks: Because of the length yeah, you have got `toFixed(6)`

stephen-ciq: If you made each number a fixed length, then yes, that comparison would work.
Meredith and I came up with a different way that she is implementing, though.
basically it's

```
compareVersions(a, b)
  if (a.majorVersion !== b.majorVersion)
    return a.majorVersion < b.majorVersion ? -1 : 1;
  if (a.minorVersion !== b.minorVersion )
    return a.minorVersion < b.minorVersion ? -1 : 1;
  if (a.patchVersion !== b.patchVersion)
    return a.patchVersion< b.patchVersion? -1 : 1;
  return 0;
```

and then you use the comparison result to know whether the whole version `a` is less than the whole version `b`.

amaroks: I think that has too many `if` `if`

```javascript
const parse = (version) => {
  return `${version.replace(/\D/g, '') + '000'}`.substr(0, 6)
}

if (parse(current) > parse(max) ) {
  return false;
}

```

meredith-ciq: @amaroks I've just checked in a refactor of the whole thing, so please take a look. 
I'd prefer not to have that many ifs either, but I've stared at your solution for several minutes and still have no idea how it works. I think it's okay to have a few extra lines to make it more readable.

amaroks: ok @meredith-ciq fair enough

gulpfile.js
```diff
@@ -51,6 +52,82 @@ const { launch, connect } = require('hadouken-js-adapter');
...
 
+	/** 
+	* Splits a string version with semantic versioning into an object with major, minor and patch versions
+	* Valid inputs are 'X.X.X' or 'vX.X.X'
+	*/
+	const createSemverObject = (version) => {
+		let tempVersionArray;
+		let semverObject;
+		if (!version || typeof version !== 'string') {
+			console.log(`Version must be type string`);
```
stephen-ciq:
```suggestion
			console.log(`Version must be type string but is ${typeof version}`);
```

gulpfile.js
```diff
@@ -51,6 +52,82 @@ const { launch, connect } = require('hadouken-js-adapter');
...
 	}
 
+	/** 
+	* Splits a string version with semantic versioning into an object with major, minor and patch versions
+	* Valid inputs are 'X.X.X' or 'vX.X.X'
+	*/
+	const createSemverObject = (version) => {
+		let tempVersionArray;
+		let semverObject;
+		if (!version || typeof version !== 'string') {
```
stephen-ciq:
```suggestion
		if (typeof version !== 'string') {
```

The `!version` check is not necessary because `typeof undefined` is `"undefined"`.  So the `typeof version !== 'string'` is all you need, since it will catch the case where `version` is undefined.

gulpfile.js
```diff
@@ -51,6 +52,82 @@ const { launch, connect } = require('hadouken-js-adapter');
...
+		semverObject = {
+			majorVersion: Number(tempVersionArray[0]) || null,
+			minorVersion: Number(tempVersionArray[1]) || null,
+			patchVersion: Number(tempVersionArray[2]) || null,
+		}
+		// If major, minor or patch versions are missing or not a number return nothing
+		if (!semverObject.majorVersion || !semverObject.minorVersion || !semverObject.patchVersion) {
+			return;
+		}
+		return semverObject
```
stephen-ciq:
```suggestion
		return semverObject;
```

gulpfile.js
```diff
@@ -51,6 +52,82 @@ const { launch, connect } = require('hadouken-js-adapter');
...
+			return;
+		}
+		// Split the version into a temp array.
+		if (version.startsWith('v')) {
+			tempVersionArray = version.split('v');
+			tempVersionArray = tempVersionArray[1].split('.');
+		} else {
+			tempVersionArray = version.split('.')
+		}
+		if (tempVersionArray)
```
stephen-ciq: Counter to what one might think, an array in Javascript is always true.

For example
```
var hmm = ''.split(',');
console.log(Boolean(hmm));
```

prints `true`.

So I think you need to check the length of the array and also have an `else` for when the check fails:
```
		if (tempVersionArray.length != 3)
    {
		// Convert each array element to a number and store in the object.
		semverObject = {
			majorVersion: Number(tempVersionArray[0]) || null,
			minorVersion: Number(tempVersionArray[1]) || null,
			patchVersion: Number(tempVersionArray[2]) || null,
		};
    }
    else
    {
         return;
    }
		// If major, minor or patch versions are missing or not a number return nothing
		if (!semverObject.majorVersion || !semverObject.minorVersion || !semverObject.patchVersion) {
			return;
		}
		return semverObject;
	}
```

meredith-ciq: Fixed that check. The error case is already handled because the function execution ends and will return automatically.

PR #499, opened 2019-11-16
Fix always on top

d4hines:
fix: #id [15774]

Accompanied by core PR: https://github.com/ChartIQ/finsemble/pull/1612

**Description of change**
This PR implements the alwaysOnTop logic layed in the [informal spec]( https://docs.google.com/document/d/1VztbpfstHT0n7xFs55igstcVi2iB43GDRP9m5nAYAsY/edit) (copied here for convenience)

> Definitions
For the purpose of this document, a “collection” is either a group or a stack. The logic below applies isomorphically to both.

>Logic
>When a window’s alwaysOnTop state is set to true, the window becomes visible and no other windows may cover that window. This state persists across workspace reloads until toggled off.
>When a window joins a collection (or creates a new collection with another window), if any of the windows in the collection are alwaysOnTop, then all of the windows become alwaysOnTop.
>When a window leaves a collection, it should return its state prior to joining the group.
>When the alwaysOnTop state of any window in a collection is toggled, every other window in the collection is also toggled.

**NOTE!** This PR only addresses groups of Web windows. Anything else is not guaranteed to work. See https://chartiq.kanbanize.com/ctrl_board/18/cards/18776/details/

**Background Info**
- launcherGroup and dockableGroup inherit from windowGroup.js.
- launcherGroups don’t deal with any AOT behavior.
- dockingMain sets up a listener for AOT changes on dockable windows; however, besides changing the AOT state of a  group, it doesn’t  do anything  with the AOT data, which should be private to the group.
- We expose an API for setting AOT; however, our own windowClient undercuts this method by using a lower level functionality. This means any change to AOT logic for single windows would have to be implemented in two places.

**Changes**
- Move AOT listener from dockingMain to windowGroup.js
  - This should have zero impact on function, and is purely about code organization
-Delete dead/unused AOT code in launcherGroup.js windowGroup.js (Sidd communicated possible plans to use the code in launcherGroup)
- Move AOT code from windowGroup.js to dockableGroup.js
- Implement the above logic in dockableGroup
   - Add new property, “prevAlwaysOnTop”, that maps window names to their AOT state prior to joining the group
    - In the “addWindow” method, if any windows in the group are AOT, set the added window to AOT
    - In the “addWindow” method, set up a listener for AOT changes that changes the AOT of every window in the group.
     - In the “removeWindow” method, 
     - According to Brad, all the windows are removed via “removeGroup” before group destruction, 
- Change windowClient to call the appropriate router query for AOT instead of calling finsembleWindow.updateOptions

In addition to implementing this logic, I made the following changes (see :
-  Removed unused `alwaysOnTop` method from windowGroup.js
- Fixed windowClient to use the router entry point for alwaysOnTop, rather than finsembleWindow's updateOptions method.

**Description of Testing**
For the following tests, a window is AOT if and only if
    - The icon is highlighted
    - It cannot be covered by a window (you'll have to test this each time
    with an Explorer/Chrome/console window).

A window is non-AOT if and only if neither of the above conditions hold.

"Toggling the group AOT" means clicking the AOT button on the leftmost
component in the group (it should be the only one visible).

If any errors appear in the CL, the tests fail. (the "Open Logger on Error" option in the CL is convenient for this).

*Test Steps*
Given Welcome components A, B, and C.
1. Set A to AOT. Snap A to B such that A is left of B. Group A and B.
    - [ ] A and B are both AOT
2. Group C with A and B.
    - [ ] C becomes AOT
3. Ungroup C
    - [ ] C returns to non-AOT
2. Toggle the group's AOT
    - [ ] Entire Group becomes non AOT
3 Disband the group
    - [ ] A is AOT, but B is not AOT.
4. Toggle A to non AOT. Group A and B again
    - [ ] Neither A nor B is AOT.
5. Cover the group and click the Toolbar's Bring to Front buttom
    - [ ] the group is brought to front
6. Cover A with another window. Click B
    - [ ] A becomes visible
7. Toggle the Group's AOT
    - [ ] A and B are still grouped and AOT
8. Reload the workspace
    - [ ] A and B are still grouped and AOT. C is non AOT.
9. Toggle the group's AOT.
    - [ ] Neither A nor B is AOT.
10. Disband the group
    - [ ] Neither A nor B is  AOT.
11. Set A to AOT. Group with B. Reload the workspace. Disband the group.
    - [ ] A is AOT, but B is non AOT.

# NOTE
This PR doesn't address groups of any windows besides web windows (i.e, stacks and native windows will be addressed in another card).

### DIFF
--- a/src-built-in/components/windowTitleBar/src/windowTitleBarComponent.jsx
+++ b/src-built-in/components/windowTitleBar/src/windowTitleBarComponent.jsx
@@ -407,7 +407,15 @@ class WindowTitleBar extends React.Component {
 				</div>
 				<div className={rightWrapperClasses} ref={this.setToolbarRight}>
 					{showDockingIcon ? <DockingButton /> : null}
-					{this.state.alwaysOnTopButton && showMinimizeIcon ? <AlwaysOnTop /> : null}
+					{/** DH 11/22/2019
+					 * Because AlwaysOnTop is a "smart" component that registers
+					 * event handlers, etc., it's not a good idea to constantly mount
+					 * and unmount it. To prevent this, we pass in a "visible" prop that,
+					 * if false, sets "display: none". Ideally, AlwaysOnTop should be a "dumb"
+					 * component, and all the event handlers, etc. should be registered in a parent component,
+					 * as this simplifies the UI and allows React to better optimize under the hood.
+					 */}
+					<AlwaysOnTop visible={this.state.alwaysOnTopButton && showMinimizeIcon}/>
 					<BringSuiteToFront />
 					{this.state.minButton && showMinimizeIcon ? <Minimize /> : null}
 					{this.state.maxButton ? <Maximize /> : null}
--- a/configs/application/components.json
+++ b/configs/application/components.json
@@ -34,7 +34,6 @@
 						"launchableByUser": true
 					},
 					"Window Manager": {
-						"alwaysOnTopIcon": false,
 						"showLinker": true,
 						"FSBLHeader": true,
 						"persistWindowState": true,
@@ -120,7 +119,6 @@
 						"launchableByUser": false
 					},
 					"Window Manager": {
-						"alwaysOnTopIcon": false,
 						"showLinker": false,
 						"FSBLHeader": true,
 						"persistWindowState": true,
@@ -149,7 +147,6 @@
 						"launchableByUser": false
 					},
 					"Window Manager": {
-						"alwaysOnTopIcon": false,
 						"showLinker": false,
 						"FSBLHeader": true,
 						"persistWindowState": true,
--- a/configs/application/components.json
+++ b/configs/application/components.json
@@ -34,6 +34,7 @@
 						"launchableByUser": true
 					},
 					"Window Manager": {
+						"alwaysOnTopIcon": false,
 						"showLinker": true,
 						"FSBLHeader": true,
 						"persistWindowState": true,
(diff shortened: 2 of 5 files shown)


src-built-in/components/windowTitleBar/src/windowTitleBarComponent.jsx
```diff
@@ -407,7 +407,15 @@ class WindowTitleBar extends React.Component {
...
-					{this.state.alwaysOnTopButton && showMinimizeIcon ? <AlwaysOnTop /> : null}
+					{/** DH 11/22/2019
+					 * Because AlwaysOnTop is a "smart" component that registers
+					 * event handlers, etc., it's not a good idea to constantly mount
+					 * and unmount it. To prevent this, we pass in a "visible" prop that,
+					 * if false, sets "display: none". Ideally, AlwaysOnTop should be a "dumb"
+					 * component, and all the event handlers, etc. should be registered in a parent component,
+					 * as this simplifies the UI and allows React to better optimize under the hood.
+					 */}
+					<AlwaysOnTop visible={this.state.alwaysOnTopButton && showMinimizeIcon}/>
```
meredith-ciq: This is a little concerning. Is there a reason alwaysOnTop can't be a dumb component? Did it work the same way before?

meredith-ciq: Answer is the work is too big to do in this effort.

d4hines: Yes, it worked the same way before, but had an intrinsic race condition. This fixes the race condition, but leaves the (sub-optimal) design in place (we can fix it in the seed refactor)

PR #506, opened 2019-11-26
15321 workspace failures

mhmcclung:

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/15321/details/)

**Description of change**
* Removed UI restriction that prevents switching workspace before the previous switch is complete. 

See corresponding Finsemble PR: https://github.com/ChartIQ/finsemble/pull/1610  --  **These two PRs must used and merged together.**

### DIFF
--- a/src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
+++ b/src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
@@ -342,7 +342,6 @@ Actions = {
 	 * Asks the user if they'd like to save their data, then loads the requested workspace.
 	 */
 	switchToWorkspace: function (data) {
-		//if (Actions.getIsSwitchingWorkspaces()) return;
 		Actions.setIsSwitchingWorkspaces(true);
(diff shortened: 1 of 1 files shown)


src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
```diff
@@ -342,7 +342,6 @@ Actions = {
 	 * Asks the user if they'd like to save their data, then loads the requested workspace.
 	 */
 	switchToWorkspace: function (data) {
-		if (Actions.getIsSwitchingWorkspaces()) return;
```
clausjs: Is there no longer a need to check this?

Garbee: Nope. Now people can be loading a workspace, decide to switch to another. We will queue the request, attempt to close the current stuff down (and cancel any pending load requests), then switch to the next workspace. If multiple are queued before the close finishes, only the last one requested is used.

PR #507, opened 2019-11-29
fix: #id [19305] correct window.options in seed components

kriswest:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/48/cards/19305/details/)

**Fix Welcome component and other example configs**
* The seed project contained some config options that were not working as they were not in the correct place (e.g. window.autoShow, which should be window.options.autoShow). This misleads clients all the time - particularly as we don't document window.options (we used to rely on OpenFin for this).

**Description of testing**
1. Run the seed project and check the Welcome component is working as normal.
1. Change some of the window.options.* settings (e.g. set `window.options.resizable = false` or `window.options.alwaysOnTop = true`) 
1. Restart Finsemble and see that the specified settings apply to the component correctly.

### DIFF
--- a/configs/application/components.json
+++ b/configs/application/components.json
@@ -8,7 +8,6 @@
 				"options": {
 					"resizable": true,
 					"autoShow": true,
-					"backgroundColor": "#22262F",
 					"alwaysOnTop": false
 				},
 				"top": "center",
--- a/configs/application/components.json
+++ b/configs/application/components.json
@@ -6,8 +6,12 @@
 				"url": "$applicationRoot/components/welcome/welcome.html",
 				"affinity": "workspaceComponents",
 				"frame": false,
-				"resizable": true,
-				"autoShow": true,
+				"options": {
+					"resizable": true,
+					"autoShow": true,
+					"backgroundColor": "#22262F",
+					"alwaysOnTop": false
+				},
 				"top": "center",
 				"left": "center",
 				"width": 400,
@@ -34,6 +38,7 @@
 						"launchableByUser": true
 					},
 					"Window Manager": {
+						"alwaysOnTopIcon": true,
 						"showLinker": true,
 						"FSBLHeader": true,
 						"persistWindowState": true,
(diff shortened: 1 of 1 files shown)


configs/application/components.json
```diff
@@ -6,8 +6,12 @@
 				"url": "$applicationRoot/components/welcome/welcome.html",
 				"affinity": "workspaceComponents",
 				"frame": false,
-				"resizable": true,
-				"autoShow": true,
+				"options": {
+					"resizable": true,
+					"autoShow": true,
+					"backgroundColor": "#22262F",
```
NinjaManatee: @kriswest Why did you add a background color?

kriswest: To stimulate discussion of what we should actually be putting here. We've had dead options in place for over a year and at present have no docs on what options are available. This is one that has been requested in the past - although it doesn't actually prevent a white flash as the component loads annoyingly. 

I'll be following up on the ticket shortly as we should be discussing which options to surface in this example (which is likely to be one of the prime ways developers know about options, whether or not we document them).

kriswest: As this doesn't actually prevent that white flash, I think it can be dropped. 

Is there anything else (in finsemble\src\common\FinsembleWindowData.d.ts) that you think would be useful to add into an explicit example @NinjaManatee?

kriswest:
```suggestion
```

configs/application/components.json
```diff
@@ -6,8 +6,12 @@
 				"url": "$applicationRoot/components/welcome/welcome.html",
 				"affinity": "workspaceComponents",
 				"frame": false,
```
kriswest:
```suggestion
```

configs/application/components.json
```diff
@@ -50,8 +53,10 @@
 				"url": "$applicationRoot/components/nonConfiguredComponent/nonConfiguredComponent.html",
 				"affinity": "systemComponents",
 				"frame": false,
```
kriswest:
```suggestion
```

PR #508, opened 2019-12-02
chore: #id 19169 Remove implementation of launchGroups in app launcher

clausjs:
chore: #id [19169]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/19169/details)

**Description of change**
* Removed app launcher's implementation of launcherGroups

Finsemble PR required to test: https://github.com/ChartIQ/finsemble/pull/1642

**Description of testing**
1. Add a `launcherGroups` property to two different component's `<name>.component` field, make the value of `launcherGroups` the same for both components.
1. Launch Finsemble
1. Launch both apps and ensure no grouping/simultaneous launching occurs
1. Make sure there is no entry in the `App Launcher` or `My Apps` menu
1. [ ] Everything behaves normally and nothing unexpected was added to the app menus

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -144,8 +144,6 @@ const { launch, connect } = require('hadouken-js-adapter');
 
 	// #endregion
 
-	// #region Script variables
-	let watchClose;
 	// If you specify environment variables to child_process, it overwrites all environment variables, including
 	// PATH. So, copy based on our existing env variables.
 	const env = process.env;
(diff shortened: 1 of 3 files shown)


clausjs:
> Pulling this out of the seed LGTM. I do recall us being asked to not delete code if we could manage it. However since this is the seed I'd rather it gone than linger to cause issues later. If we ever re-implement this it can be done through a proper layer where the core logic is down in the core.

As far as I could tell, the only thing that made this work to begin with was the code we had here in the seed. We document in Finsemble, and link to this code snippet as the example of how to implement. There is no code in the actual launcher that sets up a launch group. Instead the seed code reads the configs, and creates a launcher entry for the group. So properly 'removing' launcher groups means this code has to go

PR #512, opened 2019-12-03
feat: #id 19170 disable hyperfocus

meredith-ciq:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/19170/details)

**Description of change**
* Removed all code in the seed that called or implemented hyperfocus

**Description of testing**
1. Verified Finsemble starts
1. Verified shift-clicking on a channel no longer triggers the hyperfocus action (all other windows in a different channel minimize). It may still be treated the same as a regular left-click.


amaroks:
Looks like this PR already does some work for this card
https://chartiq.kanbanize.com/ctrl_board/18/cards/19160/details/

PR #514, opened 2019-12-03
19168 disable auto arrange

mhmcclung:
fix #id 19168

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/19168/details/)

**Description of change**
* Disable/Hide auto-arrange icon to disable auto arrange, but kept it easy to re-enable since used in testing large workspaces

**Description of testing**
Confirmed auto-arrange icon is no longer visible (when AutoArrangeEnabled = false);

### DIFF
--- a/src-built-in/components/toolbar/components/AutoArrange.jsx
+++ b/src-built-in/components/toolbar/components/AutoArrange.jsx
@@ -68,7 +68,7 @@ export default class AutoArrange extends React.Component {
 
 	render() {
 		// For 4.0 auto-arrange is disabled -- it will be re-enabled in a future release after it is improved and hardened.
-		// However, keep it easy to re-enable by setting AutoArrangeEnabled=true -- this should only be done in-house to simply testing with a large number of windows
+		// However, keep it easy to re-enable by setting AutoArrangeEnabled=true -- this should only be done for in-house testing with a large number of windows
 		const AutoArrangeEnabled = false;
 
 		if (AutoArrangeEnabled) {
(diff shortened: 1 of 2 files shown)


mhmcclung:
@Garbee the ticket said no large code blocks should be removed, so I left the code in the seed.  The seed refactor will be substantial, so I figure whoever does that can decide about auto-arrange, plus I hope we fix it soon since I've heard our field engineers say customers do use it.   As for why not just invoke the API -- that would work but I commonly use auto-arrange to make visual verification across many windows easier when stress testing, so easier to change one flag then use have a button.

mhmcclung:
changed "reveal all" button to have the window-mgmt-right as per Dan and Garbee's comments.

src-built-in/components/toolbar/components/AutoArrange.jsx
```diff
@@ -67,19 +67,29 @@ export default class AutoArrange extends React.Component {
 	}
 
 	render() {
-		const autoArrangedCss = this.state.isAutoArranged ? "auto-arranged" : "";
+		// For 4.0 auto-arrange is disabled -- it will be re-enabled in a future release after it is improved and hardened.
```
Garbee: Remove comments such as this. This repo is public and it is exposing things we may not want to.

mhmcclung: The comment documents the reality and explains the presence of the code, especially since disabling AA won't be viable for long.  But bottom line is I did remove it.

PR #538, opened 2019-12-20
Allow an installer configuration per server environment

kriswest:
fix: #20133 [CARD]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/48/cards/20133/details/)

**Description of change**
* Check if installer configs are given per environment (i.e. check for one with current environment name)
* Update default config to have a configuration per server-environment

**Description of testing**
1. Install dependencies `npm install`
1. Roll an installer for the development environment `npm run makeInstaller:dev`
1. Installer should be output into configured package directory (_/pkg-development_)
1. Installer should be named according to the per environment config: xyz-devSetup.exe
1. Desktop shortcut created by installer should be named according to the per environment config: xyz-dev


kriswest:
@clausjs 
> This will fail if NODE_ENV isn't set, perhaps there should be a default `else` case?
> 
> @kriswest this now out of sync and needs rebasing. Can you also address the above?

Do you mean in gulfile.js: makeInstaller? If so that function starts with a check on the NODE_ENV and exits with an (existing) error message if its not set, here:
https://github.com/ChartIQ/finsemble-seed/blob/e8aa7abb4daaac54b20e2146d76075c498b79ca3/gulpfile.js#L539

I merged 4.2.0 back into this - there was a small conflict in installer.js which I resolved. I did not rebase as a) last time I did that was _years_ ago b) the diff looks ok. @clausjs if that still needs doing can you point me at how you can tell and pointer on doing so (would be much appreciated)

PR #542, opened 2020-01-02
[19852] Fix: window control lost when dropping tabs

amaroks:
fix: #id [19852]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/19852/details)

**Description of change**
* Created an drop event handler for non tab areas in titlebar.
* Stop event propagation when tab is dropped on another tab.

**Description of testing**
1. Create 2-3 welcome components
1. Tab all components
1. Drag any tab and drop it on an empty area in title bar
1. [x] Shouldn't loose window control
1. [x] Should see a blue scrim when hovering on windows

See the following card to see what the bug looks like
https://chartiq.kanbanize.com/ctrl_board/18/cards/14562/details/

### DIFF
--- a/src-built-in/components/windowTitleBar/src/components/center/TabRegion.jsx
+++ b/src-built-in/components/windowTitleBar/src/components/center/TabRegion.jsx
@@ -125,6 +125,9 @@ export default class TabRegion extends React.Component {
 	 * @memberof windowTitleBar
 	 */
     stopDrag(e) {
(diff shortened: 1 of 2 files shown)


d4hines:
Does a similar change need to be made for the assimilated window floating titlebar?

amaroks:
@d4hines if they're 2 different components then probably. I will look into it.

PR #546, opened 2020-01-09
Quick components now show name in app launcher

clausjs:
fix: #id [20438]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/20438/details)

**Description of change**
* Changed AppDefinition to show name in the event that displayName is empty

**Description of testing**
1. Add a quick component through the advanced app launcher
1. [ ] The app's name is displayed in the launcher


clausjs:
> Gotta love simple conditionals. LGTM.
> 
> However, we really should have a method/getter on the app object so we don't need to remember to do this everywhere and in the same order. However, that is a design discussion that is a bit wider than fixing this one area right now.

I absolutely agree with this; hopefully the redesign will address this

PR #550, opened 2020-01-10
feat: #id 13457

JMCCROWELL:
feat: #id 13457

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/13457/details)

**Description of change**
* Created a new Welcome component with a walkthrough button
* Changed the default welcome component to newWelcome
* Changed the links to the new getting started tutorial

**Description of testing**
1. Play around with the welcome component. Is it intuitive?
1. Is the copy on the new windows correct and legible? 
1. Are the CSS and .JS files good exemplars of how to build a component in Finsemble?


NinjaManatee:
I think the content looks great! Here are some minor nitpicks on the presentation:

- It seems strange to me that "Launch Docs" opens the config reference rather than "Introduction to Finsemble".
- It might be better to add the getting started tutorial to the welcome component rather than to the Apps menu (e.g. Take a Tour, Launch Docs and Getting Started Tutorial), and launch it in a browser window like at the end of the tutorial.
- The initial size of the Workspaces tutorial window is too short. For me the min height should be 402px. 
- I think it would be good to make the tour windows singletons so you don't have multiple of the same tour window.
- I think it would be cleaner to close one window from the tour when you open the next (e.g. close the "Snapping and Docking" window after you open the "Hotkeys" window.
- I also think it would be cleaner if the windows opened at the center of the monitor (except for "Snapping and Docking"). 
- I'm leery of exposing CTRL+ALT+SHIFT+R. I think it would be better to direct them to use the file menu to restart.

kriswest:
I agree with most of Andy's comments, some thoughts:
> - I think it would be good to make the tour windows singletons so you don't have multiple of the same tour window.

I wouldn't worry about that - it won't come up if the other suggestion (closing the previous component) is followed. In fact, I'd probably go a different direction and not have a config per view. I'd probably have a single config for all of them and just override the URL when you spawn the next component, e.g.:
```
window.tourWorkspaces = function tourWorkspaces() {
	FSBL.Clients.LauncherClient.spawn("Welcome Component",
		{
			url: "$applicationRoot/components/newWelcome/tourWorkspaces.html" 
			addToWorkspace: true,
			position: "relative",
			left: 0,
			bottom: 650
		},
		function () {
		});
} 
```
This would avoid bloat in components.json (1 component rather than 4).

>- I think it would be cleaner to close one window from the tour when you open the next (e.g. close the "Snapping and Docking" window after you open the "Hotkeys" window.

Totally agree.
```
window.tourWorkspaces = function tourWorkspaces() {
	FSBL.Clients.LauncherClient.spawn("Tour Workspaces",
		{
			...
		},
		function () {
			FSBL.Clients.WindowClient.close({ removeFromWorkspace: true, closeWindow: true });
		});
}
```
Feels much tidier - otherwise, you end up manually closing 4 windows at the end (and we might want more steps in future). 

> - I also think it would be cleaner if the windows opened at the center of the monitor (except for "Snapping and Docking").

I kind of like the positions moving around some, although I would finesse the locations some:
- Tour Snap and Dock: `position: "relative", left: 150, top: 0`
- Tour Hotkeys: `position: "relative", right: 'adjacent', top: 0`
- Tour Workspaces: `position: "unclaimed", left: 'center', top: 'center'`  (or `top: 0`)

The config of the Welcome Component also needs its `top` changing to 'center'.

>- I'm leery of exposing CTRL+ALT+SHIFT+R. I think it would be better to direct them to use the file menu to restart.

Agreed - I'm leery of the fact that this particular hotkey exists :-p However, it takes quite a few modifier keys to trigger. Instead of using it have them switch back and forth from the default workspace to the new one.

- Change the PR's base to planned/4.1.0
- Update GIFs to 4.0 stylesheet
- tourWorkspaces.html has both a `<style>` tag in its header and an inline style on the Launch tutorial button, both should be moved to new classes in the CSS file. 
- the images are specific to these components, so they should be in the component's folder in src rather than in assets. If a client dumps the component the images then go with it, rather than bloating the assets directory.
- How about adding a step relating to deploying your finsemble desktop, a link to config reference and then finally pointing them to [support@finsemble.com](mailto:support@finsemble.com)?

PR #556, opened 2020-01-17
20492 fdc3 combined searching

jason-ciq:
fix: #id [20492]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/20492/details)

**Tag search now queries FDC3 server**
* Tag search used to just filter locally, breaking combinatory searches. Tags will now be sent to the appd server for searching.
* SearchText was moved to the distributed store. We need to know this in several places.
* goHome() was placed into store actions to properly react to combinatory searches.

**Description of testing**

1. Build and set up app catalog
1. Perform searches using tags and text. 
1. [x] Combinatory searches work properly, and when both tags and search text are cleared, app catalog returns to home.


src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -74,36 +85,24 @@ function getToolbarStore(done) {
 	});
 }
 
+function goHome() {
```
Garbee: Please add a jsdoc for this function.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -112,60 +111,38 @@ function _addActiveTag(tag) {
...
+
+function filterApps() {
+	let { activeTags, searchText } = data;
+	
+
+	if (activeTags.length === 0 && searchText === "") {
+		goHome();
+	} else {
+		appd.search({ text: searchText, tags: activeTags }, (err, data) => {
+			if (err) console.log("Failed to search apps");
```
Garbee: So, there was an error but we're going to continue to work with `data` that could be unusable information?

I'd expect a different exit pathway for error handling compared to continuing to work. If we can continue operations just fine, why log this at all?

Even if we do need this log, it should probably fall towards a warning. And also be logged to the Central Logger and not the DevTools console.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -112,60 +111,38 @@ function _addActiveTag(tag) {
...
 function _clearActiveTags() {
-	getStore().setValue({
-		field: "activeTags",
-		value: []
-	});
+	getStore().setValue({ field: "activeTags", value: [] });	
+}
+
+
+function filterApps() {
```
Garbee: JSDoc please.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -112,60 +111,38 @@ function _addActiveTag(tag) {
...
-		value: newActiveTags
-	}, {
-		field: "filteredApps",
-		value: newApps
-	}]);
+	getStore().setValue({ field: "activeTags", value: data.activeTags	});
+	filterApps();
 }
 
 function _refreshTags() {
```
Garbee: If all this is now doing is calling `filterApps` why don't we just update all calls to this to go straight to that and delete this?

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -383,6 +360,27 @@ function clearFilteredApps() {
 	});
 }
 
+function setSearchValue(val) {
```
Garbee: JSDoc please.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -383,6 +360,27 @@ function clearFilteredApps() {
...
+	getStore().setValue({field: "searchText", value: val})
+}
+
+function getSearchValue() {
```
Garbee: JSDoc please.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -383,6 +360,27 @@ function clearFilteredApps() {
...
+	return data.searchText;
+}
+
+function clearSearchText() {
```
Garbee: JSDoc please.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -383,6 +360,27 @@ function clearFilteredApps() {
...
+	getStore().setValue({field: "searchText", value: ""})
+}
+
+function getForceSearch() {
```
Garbee: JSDoc please.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -383,6 +360,27 @@ function clearFilteredApps() {
...
+	return data.forceSearch;
+}
+
+
+function setForceSearch(val) {
```
Garbee: JSDoc please.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
jason-ciq: Reasonable. I'll take it out.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
jason-ciq: This was how it was previously written. However, I agree. Will log to central logger as an error and return.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -74,98 +85,67 @@ function getToolbarStore(done) {
...
+function filterApps() {
+	let { activeTags, searchText } = data;
+	
+
+	if (activeTags.length === 0 && searchText === "") {
+		goHome();
+	} else {
+		appd.search({ text: searchText, tags: activeTags }, (err, data) => {
+			if (err) {
+				Logger.system.error("FDC3 App search failed!: ${err}");
```
Garbee: (Clippy Voice): It looks like you're trying to interpolate a variable into a string. In JavaScript this should be done with back-ticks and not double quotes.

jason-ciq: Oops! Good catch. This has been resolved.

PR #558, opened 2020-01-17
18740 linker ui refactor - REVIEW - DRAFT ONLY

carterbs:

**Description of change**
* Changes

**Description of testing**
1. Step 1.
1. Step 2.
1. [x] Thing happened.
1. [ ] Thing did not happen.
1. Step 3


src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -0,0 +1,184 @@
...
+        case TOGGLE_CHANNEL_REQUEST:
+            const newState_request: Linker = {
+                ...state,
+                processingRequest: true
+            };
+
+            const cmd = Cmd.run(linkChannel, {
+                successActionCreator: () => toggleSuccess(payload.channelID),
+                failActionCreator: () => toggleFailure(),
+                args: [newState_request.channels[payload.channelID].name, newState_request.channels[payload.channelID].active, newState_request.windowIdentifier]
```
carterbs: i really don't like these args. Please declare variables instead of accessing all of these values. What if newState_request.channels === undefined? All this stuff would barf.

src-built-in/components/UIAPI/types.ts
```diff
@@ -0,0 +1,36 @@
+export interface RootState {
+    linker: Linker
+}
+
+export interface Linker {
```
carterbs: docs

src-built-in/components/linker/src/linker.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    useEffect(() => {
+		dispatch(linkerActions.init());
+        return () => {
+            dispatch(linkerActions.cleanUp());
+        }
+	}, []);
+
+	const allChannels = Object.values(linker.channels);
+    const channelsElements = allChannels.map(({color, name, active, id}) => {
```
carterbs:
```suggestion
    const channelElements = allChannels.map(({color, name, active, id}) => {
```

src-built-in/components/linker/src/linker.tsx
```diff
@@ -0,0 +1,75 @@
...
+        const groupClass = `linkerGroup ${color}`;
+        const style = {
+            backgroundColor: color,
+            border: `1px solid ${color}`
```
carterbs: what if color is undefined?

src-built-in/components/UIAPI/actions/linkerActions.ts
```diff
@@ -0,0 +1,79 @@
...
+    TOGGLE_CHANNEL_SUCCESS, 
+    TOGGLE_CHANNEL_FAILURE,
+    LINKER_INIT,
+    LINKER_INIT_SUCCESS,
+    LINKER_CLEANUP,
+    UPDATE_ACTIVE_CHANNELS
+} from "../actionTypes";
+
+// Functions signature to expose to the client
+const toggleChannel = (channelID: number) => {
```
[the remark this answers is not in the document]
thorsent: Would like a description for each action.

src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -0,0 +1,184 @@
+import { loop, Cmd } from 'redux-loop';
+import {
```
thorsent: Namespaces should eliminate all this importing.

src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -0,0 +1,184 @@
...
+        if (!isActive) {
+            console.log("linking the window: ", windowIdentifier);
+            FSBL.Clients.LinkerClient.linkToChannel(channelName, windowIdentifier, callback);
+        } else {
+            FSBL.Clients.LinkerClient.unlinkFromChannel(channelName, windowIdentifier, callback);
+        }
+    });
+}
+
+function initializeLinker() {
```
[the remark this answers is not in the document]
thorsent: Agree. Would recommend moving window management into local functions.

src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -0,0 +1,184 @@
...
+            return loop(state, Cmd.run(initializeLinker, {
+                successActionCreator: initSuccess,
+            }));
+        case LINKER_INIT_SUCCESS:
+            const newState = payload.value;
+            return loop(newState, Cmd.run(() => FSBL.Clients.WindowClient.fitToDOM()));
+        case TOGGLE_CHANNEL_REQUEST:
+            const newState_request: Linker = {
+                ...state,
+                processingRequest: true
```
thorsent: @stephen-ciq @mesner I'm not sold on the extra code necessary to maintain a "processingRequest" state.

src-built-in/components/UIAPI/reducers/linker.ts
thorsent: Is it [realistically] possible for newState_request.channels to be undefined though? A developer would need to write code to actively delete the channels member. That seems highly unlikely in unidirectional flow.

src-built-in/components/UIAPI/reducers/linker.ts
thorsent: It presupposed that a linker couldn't have two active requests. I could easily see a situation where an end user would toggle more than one color at a time.

src-built-in/components/linker/src/linker.tsx
```diff
@@ -0,0 +1,75 @@
...
+    });
+
+    return (
+        <div className="linkerContainer">
+            {channelsElements}
+        </div>
+    );
+}
+
+if (window.FSBL && FSBL.addEventListener) {
```
thorsent: @stephen-ciq Would like to see this finsemble boilerplate moved inside the Provider.

amaroks:
@zona-ciq I wrote tests for both action creators and reducer. I covered pretty much all testable methods. However, there 4 pending reducer tests (where redux-loop is used). Some changes needs to be made in order to make these 4 cases testable.

- We need to make sure that all methods passed to `cmd.run()` referenceable in order to be able to spy/stub them.
- We probably should not use arrow functions inside `cmd.run()`.
- All calls to Finsemble API `FSBL.xx` need to be wrapped in functions.

For an example:
`javascript    
 return loop(newState, Cmd.run(() => FSBL.Clients.WindowClient.fitToDOM()));
`
Is not testable because when you would call the reducer within the unit test, the expected state must be wrapped in `redux-loop`'s loop method and also pass the same `cmd.run(someFunction)`

Assertion example for `redux-loop`:
```javascript
  assert.deepEqual(reducerOutput, loop(
    { loading: true },
    Cmd.run(someFunction, {
      successActionCreator: someActionCreator,
      failActionCreator: anotherActionCreator,
      args: [1]
    })
  ));
```
Note that all the methods passed to `cmd.run` must be somehow abstracted and somehow referenceable.

src-built-in/components/linker/src/linker.tsx
```diff
@@ -0,0 +1,75 @@
...
+import * as ReactDOM from "react-dom";
+
+import "./css/linkerWindow.css";
+import "../../../../assets/css/font-finance.css";
+import "../../../../assets/css/finsemble.css";
+import { RootState } from '../../UIAPI/types';
+import store from '../../UIAPI/store';
+import * as linkerActions from "../../UIAPI/actions/linkerActions";
+
+declare var FSBL: any;
```
amaroks: I think we're doing this in many places. Might be a good idea to move them to a ts declaration file to avoid repeating.

src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -0,0 +1,184 @@
...
+		finsembleWindow.focus();
+    });
+    
+    let nextChannelId: number = 0;
+    const linkerInitState: Linker = Object.assign({}, initialState);
+    const initialChannels: Channels = {};
+    const initialNametoId: NameToId = {};
+
+    return new Promise((res, rej) => {
+        FSBL.Clients.LinkerClient.getAllChannels().forEach((channel: Channel) => {
```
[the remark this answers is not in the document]
zona-ciq: Agreed. Done

src-built-in/components/UIAPI/reducers/linker.ts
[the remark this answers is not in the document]
zona-ciq: 👍  Done.

src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -0,0 +1,184 @@
...
+        linkerInitState.channels = initialChannels;
+        linkerInitState.nameToId = initialNametoId;
+
+        FSBL.Clients.RouterClient.addResponder("Finsemble.LinkerWindow.SetActiveChannels", function (err: any, msg: any) {
+            if (err) {
+                return rej(`Failed to add Finsemble.LinkerWindow.SetActiveChannels Responder: ${err}`);
+            }
+            const activeChannels = msg.data.channels;
+
+            const activeChannelIds: any[number] = [];
```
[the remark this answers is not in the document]
zona-ciq: For now it is just a number, which I think is sufficient to identify a linker channel

src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -0,0 +1,184 @@
...
+            activeChannels.forEach((channel: Channel) => {
+                activeChannelIds.push(linkerInitState.nameToId[channel.name]);
+            });
+            const updatedChannels: Channels = Object.assign({}, linkerInitState.channels);
+            activeChannelIds.forEach((channelId: number) => {
+                updatedChannels[channelId].active = true;
+            });
+
+            linkerInitState.windowIdentifier = msg.data.windowIdentifier;
+            FSBL.Clients.Logger.system.log("toggle Linker window");
```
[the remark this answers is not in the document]
zona-ciq: This is just from the original logging. I will add log messages in the end of this refactor

src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -0,0 +1,184 @@
...
+            msg.sendQueryResponse(null, {});
+            store.dispatch(updateActiveChannels(msg.data));
+            
+            FSBL.Clients.ConfigClient.getValue("finsemble.accessibleLinker", (err: any, value: boolean) => {
```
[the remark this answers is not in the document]
zona-ciq: I don't think you will think this is necessary anymore after I refactored the initializeLinker function. The callback only contains 2 commands now

src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -0,0 +1,184 @@
...
+                if (err) {
+                    rej(`Error getting accessibleLinker value: ${err}`);
```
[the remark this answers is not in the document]
zona-ciq: If we choose to we can add additional behavior to address this failure in the 'LINKER_INIT_FAILURE' case, for now I will add a console error message and when I add additional log messages to the linker code I will send a message to central logger as well.

src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -0,0 +1,184 @@
...
+// The linker's reducer
+const linker = (state = initialState, action: LinkerAction) => {
+    const { type, payload } = action;
+    switch (type) {
+        case LINKER_INIT:            
+            return loop(state, Cmd.run(initializeLinker, {
+                successActionCreator: initSuccess,
+            }));
+        case LINKER_INIT_SUCCESS:
+            const newState = payload.value;
```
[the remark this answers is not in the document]
zona-ciq: this value is sent by the previous "LINKER_INIT" case's effectful code, which is sent from the linker service. When we call `res(newLinkerState)`, this value is passed down through payload.value. Here we are not altering the original state, but simply getting the new state values from the payload.

src-built-in/components/UIAPI/reducers/linker.ts
zona-ciq: The state would be initialized when the component first mounts and we call `linkerActions.init()`. If the application ever gets to this code it means we have already successfully initialized the linker. However, we do need to discuss what we should do with the init failure case, for now it does nothing other than some logging messages. The linker code before the refactor didn't address this and I'm simply modeling the previous behavior.

src-built-in/components/linker/src/linker.tsx
zona-ciq: If the user configures any linker channel, they need to specify its name and color in the config. The original code is not validating this color value so I didn't do anything here. If you think we should specify a default color I can consult Dan Nicolai on which color to use.

src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -0,0 +1,184 @@
...
+            });
+            const updatedChannels: Channels = Object.assign({}, linkerInitState.channels);
+            activeChannelIds.forEach((channelId: number) => {
+                updatedChannels[channelId].active = true;
+            });
+
+            linkerInitState.windowIdentifier = msg.data.windowIdentifier;
+            FSBL.Clients.Logger.system.log("toggle Linker window");
+            msg.sendQueryResponse(null, {});
+            store.dispatch(updateActiveChannels(msg.data));
```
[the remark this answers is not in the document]
zona-ciq: When we initialize the component we would send 2 updates to the linker state with the same payload. First is through `store.dispatch`, second is through `res(newLinkerState)`. The store will update twice with the same content, so it wouldn't cause any issues.

We need the `store.dispatch` there because we're listening on any channel updates when user switches the linker window for different components so that we can dispatch an action to update the active channels. `res(newLinkerState)` wouldn't be invoked after the initial linker initialization. 

Please let me know if this is still unclear, I can explain in person.

src-built-in/components/linker/src/linker.tsx
stephen-ciq: @thorsent I apologize for just now getting to this.

in a short convo with Zona about how to fulfill your request, the best thing I could come up with was this pseudocode.  it puts the logic of "is fsbl ready" in the store, which is provided by the provider, and no component other than the FSBLReadyGuard has to worry about it.  please provide feedback as to whether this meets your expectations when you get a chance =)

```
function addFSBLReadyListener ()
{
  if (window.FSBL && FSBL.addEventListener) {
	FSBL.addEventListener("onReady", FSBLReady);
  } else {
	window.addEventListener("FSBLReady", FSBLReady)
  }
}

function FSBLReducer(state, action)
{
switch (action.type)
case init:
  return loop(state, Cmd.run(addFSBLReadyListener, ...);
case FSBLReady:
  state.fsblIsReady = action.payload.fsblIsReady;
...
}

function FSBLReadyWrapper(props.children) {
    fsblReadySelector = useSelector(fsblIsReady );
    if (fsblReadySelector())
      // return props.children
    else
     // return "Loading" or <div /> or whatever
}
function FSBLReady() {
	ReactDOM.render(
	<Provider store={store}>
            <FSBLReadyGuard >
    		    <LinkerRefactored />
            </FSBLReadyGuard >
  	</Provider>,
   	document.getElementById("main")
   	);
}
```

src-built-in/components/UIAPI/reducers/linker.ts
stephen-ciq: Good point.  @zona I think I agree with Terry.  It is a judgment call, but Terry and Ian have been clear in their design requirement for context-dependent actions.  So because of that, for things like linking which should happen very quickly, I think it's totally fine to not have a "processingRequest" flag in the state and explicitly document why.

stephen-ciq: and also, maintaining the processingRequest for each link action is more work as Terry points out, so that's another reason to just not do it at all, in order to make the code easier to read.  the downside is that it's technically possibly for race conditions to occur, but we are explicitly making that tradeoff.

src-built-in/components/linker/src/linker.tsx
stephen-h-d: @zona-ciq, @mesner mentioned to me that there are library facilities for async initialization that would help make this pseudocode a reality/clean it up and make it clearer.  I'd suggest looking at those/chatting with him.

src-built-in/components/UIAPI/reducers/linker.ts
carterbs: I vote to keep it. Forces us to talk about the failure state.

src-built-in/components/UIAPI/reducers/linker.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+            payload: {}
+        };
+        const output = reducer(undefined, action);
+        assert.deepEqual(output, initialState);
+    });
+    
+    // Init request shouldn't change any state value
+    it('Should handle init', () => {
+        const output: any = reducer(initialState, actions.init());
+        assert.deepEqual(initialState, output[0]);
```
amaroks:
```suggestion
       expectedOutput = {
         loop(initialState, Cmd.run(initializeLinker, {
                successActionCreator: initSuccess,
            }));
       }
        assert.deepEqual(expectedOutput, output);
```

PR #564, opened 2020-01-21
20921 app catalog back button

jason-ciq:
fix: #id [20921]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/20921/details)

**Cleaned up state/store logic for App Catalog**
We had several listeners that were copying data to and from state. When we moved the state data into store for combinatory searching, we (I) introduced some timing issues with those listeners.

* ActiveApp, filteredApps, activeTags... Are all moved to central store entirely. We should not be relying on copies of these in state.
* Listeners have been cleaned up to prevent the exchange of local state to store. 
* Helpers have been added to get store values for UI instead of relying on state data.

**Description of testing**

1. Build this branch with Advanced App catalog: https://app.getguru.com/card/i8gLqdpT/Setting-Up-Advanced-App-Catalog
1. Verify that you can search via any combination of tags and text.
1. Verify that when no tags are selected and the search field is empty, AAC returns to the home page.
1. Verify that the "Back" button will return you to the home page under all search conditions
1. Verify that you can click on a component title and bring up the showcase.
1. Verify that the "Back" button will take you back to the homepage from the showcase.

### DIFF
--- a/src-built-in/components/advancedAppCatalog/src/app.jsx
+++ b/src-built-in/components/advancedAppCatalog/src/app.jsx
@@ -24,20 +24,16 @@ export default class AppMarket extends React.Component {
 		this.state = {
 			apps: [],
 			isLoading: false,
-			filteredApps: [],
 			serverError: false,
 			installed: [],
 			tags: [],
-			activeTags: [],
-			activeApp: null,
-			//forceSearch: false,
 			installationActionTaken: null
 		};
 		this.bindCorrectContext();
 	}
 	bindCorrectContext() {
 		this.addedAppsChanged = this.addedAppsChanged.bind(this);
-		this.filteringApps = this.filteringApps.bind(this);
+		this.update = this.update.bind(this);
 		this.goHome = this.goHome.bind(this);
 		this.addTag = this.addTag.bind(this);
 		this.removeTag = this.removeTag.bind(this);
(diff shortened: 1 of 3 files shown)


src-built-in/components/advancedAppCatalog/src/app.jsx
```diff
@@ -173,6 +168,47 @@ export default class AppMarket extends React.Component {
...
+	}
+
+	/**
+	 * Checks if we have an activeApp defined
+	 *
+	 * @returns {boolean}
+	 * @memberof AppMarket
+	 */
+	isActiveApp() {
+		return storeActions.getActiveApp() !== null;
```
amaroks:
```suggestion
		return storeActions.getActiveApp();
```
I don't think you need `!== null` here. The `getActiveApp()` method will either return an active app or nothing (which can be evaluated as Boolean)

src-built-in/components/advancedAppCatalog/src/app.jsx
```diff
@@ -255,7 +290,8 @@ export default class AppMarket extends React.Component {
 		});
 	}
 	getPageContents() {
-		let { filteredApps, activeTags } = this.state;
+		let filteredApps = this.getFilteredApps();
```
amaroks:
```suggestion
		const filteredApps = this.getFilteredApps();
```

src-built-in/components/advancedAppCatalog/src/app.jsx
```diff
@@ -255,7 +290,8 @@ export default class AppMarket extends React.Component {
...
+		let activeTags = this.getActiveTags();
```
amaroks:
```suggestion
		const activeTags = this.getActiveTags();
```

src-built-in/components/advancedAppCatalog/src/app.jsx
jason-ciq: I replicated the logic that was in the component. Really, if what you're saying is the case, then this entire function is superfluous as we could just call getActiveApp().

Stylistically, when the difference is this small, I prefer to be more declarative in what the code is doing instead of relying on assuming the system will do something. The intent is clearer here for the next programmer.

PR #581, opened 2020-02-04
fix: #id 21548If manifest from appd is an object, use it

clausjs:
fix: #id [21548]

**Description of change**
* If app.manifest is an object, use it instead of custom manifest when adding an app to the launcher

**Description of testing**
1. Remove welcome component from the components.json.
2. Clear cache
3. Add the following code to the 'manifest' field of the Welcome component in data.json of the appd:
`'window': {
				'url': '$applicationRoot/components/welcome/welcome.html',
				'affinity': 'workspaceComponents',
				'frame': false,
				'resizable': true,
				'autoShow': true,
				'top': 'center',
				'left': 'center',
				'width': 400,
				'height': 432
			},
			'component': {
				'displayName': 'Welcome Component',
				'spawnOnStartup': false,
				'preload': '$applicationRoot/preloads/zoom.js'
			},
			'foreign': {
				'services': {
					'windowService': {
						'allowSnapping': true,
						'canGroup': true,
						'allowAutoArrange': true,
						'allowMinimize': true
					}
				},
				'components': {
					'App Launcher': {
						'launchableByUser': true
					},
					'Window Manager': {
						'showLinker': true,
						'FSBLHeader': true,
						'persistWindowState': true,
						'title': 'Welcome to Finsemble'
					},
					'Toolbar': {
						'iconClass': 'ff-component'
					}
				}
			}`
4. Launch Finsemble with the Advanced App Launcher/Advanced App Catalog configured
5. Add the Welcome Component from the app catalog
6. Zoom preload should work (Ctrl +, Ctrl -)

### DIFF
--- a/src-built-in/components/advancedAppCatalog/src/components/Showcase/AppDescription.jsx
+++ b/src-built-in/components/advancedAppCatalog/src/components/Showcase/AppDescription.jsx
@@ -13,7 +13,6 @@ const DEFAULT_APP_DESCRIPTION = "Get started by adding this component to Finsemb
  */
 const AppDescription = props => {
 
-	console.log(props);
 	const description = props.description === undefined ? DEFAULT_APP_DESCRIPTION : props.description;
 
 	return (
--- a/src-built-in/components/advancedAppCatalog/src/components/Showcase/AppDescription.jsx
+++ b/src-built-in/components/advancedAppCatalog/src/components/Showcase/AppDescription.jsx
@@ -10,11 +10,14 @@ import React from "react";
  * @param {string} props.description The app description
  */
 const AppDescription = props => {
+
+	const description = props.description === null ? "Get started by adding this component to Finsemble now!" : props.description;
+
 	return (
 		<div className="app-notes description">
 			<span className="showcase-label">Description</span>
 			<div className="description-content">
-				<div>{props.description || DEFAULT_APP_DESCRIPTION}</div>
+				<div>{description}</div>
 			</div>
 		</div>
 	);
--- a/src-built-in/components/advancedAppCatalog/src/components/Showcase/AppDescription.jsx
+++ b/src-built-in/components/advancedAppCatalog/src/components/Showcase/AppDescription.jsx
@@ -4,6 +4,8 @@
 */
 import React from "react";
 
+const DEFAULT_APP_DESCRIPTION = "Get started by adding this component to Finsemble now!";
+
 /**
  * AppShowcase description section
  * @param {object} props Component props
@@ -11,7 +13,8 @@ import React from "react";
  */
 const AppDescription = props => {
 
-	const description = props.description === null ? "Get started by adding this component to Finsemble now!" : props.description;
+	console.log(props);
+	const description = props.description === undefined ? DEFAULT_APP_DESCRIPTION : props.description;
 
 	return (
 		<div className="app-notes description">
--- a/src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
+++ b/src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
@@ -178,6 +178,25 @@ async function getTags() {
 }
 
 /**
+ * @param {string} data The data to attempt to parse into JSON
+ * @param {string} title The name of the app that is being parsed
+ * @param {boolean} logError Whether or not to log an error if the parsing fails
+ * @private
+ */
+function tryToParseJSON(data, title, logError) {
+	let output;
+	try {
+		output = JSON.parse(data);
+	} catch(e) {
+		if (logError) {
+			FSBL.Clients.Logger.system.warn(`Failure encountered parsing manifest for ${title}. Creating a default manifest`);
+		}
+	} finally {
+		return output;
+	}
+} 
+
+/**
  * Function to "install" an app. Adds the id to a list of installed apps
  * @param {string} name The name of the app
  */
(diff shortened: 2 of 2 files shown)


NinjaManatee:
@clausjs The AppD specification says that manifest is a string, not an object. Shouldn't we conform to the specification?
>manifest required | stringURI or full JSON of the application manifest providing all details related to launch and use requirements as described by the vendor. The format of this manifest is vendor specific, but can be identified by the manifestType attribute.

src-built-in/components/advancedAppCatalog/src/components/Showcase/AppDescription.jsx
```diff
@@ -10,11 +10,14 @@ import React from "react";
  * @param {string} props.description The app description
  */
 const AppDescription = props => {
+
+	const description = props.description === null ? "Get started by adding this component to Finsemble now!" : props.description;
```
NinjaManatee: Do we want the default string as a magic string or as a constant?

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -235,10 +235,16 @@ async function addApp(id, cb = Function.prototype) {
 		delete appConfig.window.url;
 	}
 
-	if (typeof app.manifest !== "object") {
-		appConfig.manifest = { ...appConfig };
+	let manifest;
+	try {
+		manifest = JSON.parse(app.manifest);
+	} catch(e) {
+		FSBL.Clients.Logger.warn(`${app.title || app.name} is missing a valid manifest. Creating a default manifest`);
```
NinjaManatee: Should this be a system warning?

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -235,10 +235,16 @@ async function addApp(id, cb = Function.prototype) {
 		delete appConfig.window.url;
 	}
 
-	if (typeof app.manifest !== "object") {
-		appConfig.manifest = { ...appConfig };
+	let manifest;
+	try {
+		manifest = JSON.parse(app.manifest);
```
Garbee: If the `app` has already been parsed, `app.manifest` should be an object, which `JSON.parse` will fail against even though the manifest is as defined by the server.

Are we absolutely certain that only a string representation of their manifest will end up here?

clausjs: Te spec defines the prop as a string. If a user's appD server delivers anyting but a string for this prop, they would be going against the spec, and expect a failure. There definitely needs to be an entirely different error if the JSON is not valid, but validation is on my roadmap

clausjs:
> Should this console log stay in the code?

Nope, I missed it when I cleaned up. Good call

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
Garbee: Talked directly with Joe on this again. I was confused based on the tests I saw yesterday. Turns out when testing he initially had invalid FDC3 structure to mess with the logic here. It was after that he learned it shouldn't be an object but a string by the specification.

All good here.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -235,8 +235,28 @@ async function addApp(id, cb = Function.prototype) {
...
+		} catch(e) {
+			try {
+				//If parsing fails, assume the manifest is a valid url that will return JSON
+				const urlRes = await fetch(app.manifest, { method: "GET" });
+
+				//Attempt to parse the url response
+				manifest = await urlRes.json();
+			} catch(e) {
+				FSBL.Clients.Logger.system.warn(`${app.title || app.name} is missing a valid manifest or URI that delivers a valid JSON manifest. Unable to add app`);
+				return cb();
```
Garbee: Here and on 259 there is an error state yet the callback is being executed with no context of an error. *Something* should be passed up to let the caller know this failed.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -235,8 +235,28 @@ async function addApp(id, cb = Function.prototype) {
...
+	if (app.manifestType === "Finsemble") {
+		try {
+			//Attempt to parse a string as JSON
+			manifest = JSON.parse(app.manifest);
+		} catch(e) {
+			try {
+				//If parsing fails, assume the manifest is a valid url that will return JSON
+				const urlRes = await fetch(app.manifest, { method: "GET" });
+
+				//Attempt to parse the url response
```
Garbee: Nit: Unneeded comment.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -235,8 +235,28 @@ async function addApp(id, cb = Function.prototype) {
 		delete appConfig.window.url;
 	}
 
-	if (typeof app.manifest !== "object") {
-		appConfig.manifest = { ...appConfig };
+	let manifest;
+	if (app.manifestType === "Finsemble") {
+		try {
+			//Attempt to parse a string as JSON
```
Garbee: Nit: Unneeded comment

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
```diff
@@ -235,8 +235,28 @@ async function addApp(id, cb = Function.prototype) {
...
+			manifest = JSON.parse(app.manifest);
+		} catch(e) {
+			try {
+				//If parsing fails, assume the manifest is a valid url that will return JSON
+				const urlRes = await fetch(app.manifest, { method: "GET" });
+
+				//Attempt to parse the url response
+				manifest = await urlRes.json();
+			} catch(e) {
+				FSBL.Clients.Logger.system.warn(`${app.title || app.name} is missing a valid manifest or URI that delivers a valid JSON manifest. Unable to add app`);
```
Garbee: Add period at the end of "Unable to add app". Without a defined language structure to follow let's at least try to be consistent. You don't have punctuation at the end here where on the warning in 258 you do.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
Garbee: So, just mentioned this one to Joe on Slack. This is a pre-existing condition callback condition. As much as it is terrible to keep letting it happen, right now we shouldn't hold this up on the API being crud.

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
clausjs: I'm going to leave these comments, since it is not obvious why I would be doing these two things on a string property. If anything, they should be fleshed out so the next person to come along and look at this code doesn't need to have a comprehensive understanding of the FDC3 spec

src-built-in/components/advancedAppCatalog/src/stores/storeActions.js
clausjs: I fleshed out this comment more

PR #582, opened 2020-02-05
41445 multiple workspace prompts

mhmcclung:
fix: #id 21445

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/21445/details/)

**Description of change**
* Allow only one workspace prompt to be displayed, locking out workspace switches until the prompt has been responded to
* Fixed problem in workspaceManagementMenuStore that prevented async.waterfall(tasks, Actions.onAsyncComplete) from finishing -- switchWorkspace didn't have a callback

**Description of testing**
1. Create a workspace pin and switch to that workspace
1. Dirty the workspace by moving a window
1. Click on workspace pin multiple time.  The first time a "save" prompt should be display, but not for the other times.  Can confirm only one prompt by responding to it (you won't find others beneath it).


src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
```diff
@@ -350,11 +363,12 @@ Actions = {
 		 * Actually perform the switch. Happens after we ask the user what they want.
 		 *
 		 */
-		function switchWorkspace() {
+		function switchWorkspace(callback) {
```
Garbee: Why is the callback needing to be introduced?

If it is required, JSDoc needs updating.

src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
```diff
@@ -350,11 +363,12 @@ Actions = {
...
 			FSBL.Clients.WorkspaceClient.switchTo({
 				name: name
 			}, () => {
-				Actions.setIsSwitchingWorkspaces(false);
+					Actions.setIsSwitchingWorkspaces(false);
+					callback();
```
Garbee: The callback wasn't needed before to call this, now it's being added as a non-optional parameter. If this is to be added it should be optional and not fail if unprovided.

If it is to be required, we need to make sure any calls to it we make are up-to-date. Since it is on the front-end though, this does expose a breakage concern.

src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
mhmcclung: This change fixes a different bug that was indirectly causing my fix to not work.  switchWorkspace is an internal function passed into async.waterfall(), so it has to have a callback.  Without the callback async.waterfall never finished and the corresponding "done" function (Actions.onAsyncComplete) was never called.  So this change just fixes preexisting code generally unrelated to my change.

src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
mhmcclung: See the above comment.  It was needed before -- this just never worked correctly.  The callback is a non-optional parameter since passed into async.waterfall().

src-built-in/components/workspaceManagementMenu/src/stores/workspaceManagementMenuStore.js
Garbee: Ok, just add a JSDoc for it and that's reasonable. Just making sure since it is changing the signature of the function.

mhmcclung: done

PR #585, opened 2020-02-05
feat: #id 21677 dock and drag toolbar

carterbs:
Add dragHandle to the toolbar shell.
Remove the old SVG.
Add it inline instead.


src-built-in/components/toolbar/src/app.jsx
```diff
@@ -5,16 +5,17 @@
...
 import onReady from "../../UIAPI/effects/index";
 import ReactDOM from "react-dom";
 import React from "react";
-
+import DragHandle from '../../toolbarRefactored/DragHandle';
 const ToolbarShell = () => {
 	return (
 		<ToolbarRefactored>
-			New Toolbar. Delete this line when the first component is added.
+			<DragHandle/>
```
amaroks:
```suggestion
			<DragHandle />
```

PR #586, opened 2020-02-06
feat: #id 21621 active workspace hook

carterbs:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/21621/details)

**Description of change**
* Added a hook to retrieve and set the active workspace name.
* Added code to dispatch the set_active_workspace_name action.
* Added a reducer for workspace state.
* Added unit tests for the very basic reducer.

**Description of testing**
1. If you look at the first commit on this PR, you'll see a little component I made called `<Aws/>`. It uses the hook to get the active workspace, and when you click it, it dispatches an action to change the activeworkspace name. That component then updated with the activeworkspace name, showing that the code worked.
1. Ran unit tests.


src-built-in/components/UIAPI/hooks/workspaces.ts
```diff
@@ -0,0 +1,38 @@
+import { useEffect } from "react";
+import { useDispatch, useSelector } from 'react-redux';
+import { RootState, WorkspaceState } from '../types';
+import * as Actions from '../actions/workspaceActions';
+
+/**
+ * A hook for getting the activeWorkspaceName,
+ * and setting the activeWorkspaceName.
```
d4hines: You can't set the active workspace name this way.

src-built-in/components/UIAPI/hooks/workspaces.ts
```diff
@@ -0,0 +1,38 @@
...
+	useEffect(() => {
+		const setInitialActiveWorkspace = async () => {
+			const { data: aws } = await FSBL.Clients.WorkspaceClient.getActiveWorkspace();
+			setActiveWorkspaceName(aws.name);
+		}
+		setInitialActiveWorkspace();
+	}, []);
+
+	return {
+		activeWorkspaceName: activeWorkspace.name,
```
d4hines: Why wouldn't we expose the whole active workspace?

src-built-in/components/UIAPI/hooks/workspaces.ts
carterbs: removed

src-built-in/components/UIAPI/hooks/workspaces.ts
```diff
@@ -0,0 +1,54 @@
...
+			const { data: aws } = await FSBL.Clients.WorkspaceClient.getActiveWorkspace();
+			setActiveWorkspace(aws);
+		}
+		// When the workspace updates, update the active workspace name.
+		const listenForWorkspaceUpdates = () => {
+			return FSBL.Clients.RouterClient.subscribe("Finsemble.WorkspaceService.update", onWorkspaceUpdate);
+		}
+
+		setInitialActiveWorkspace();
+		const WorkspaceUpdateSubscribeID = listenForWorkspaceUpdates();
```
sgd2z: I dont understand the point of this one line function and what you are getting beyond this:
```
const WorkspaceUpdateSubscribeID = FSBL.Clients.RouterClient.subscribe("Finsemble.WorkspaceService.update", onWorkspaceUpdate);
```

carterbs: Made the body of useEffect look cleaner.

src-built-in/components/UIAPI/hooks/workspaces.ts
```diff
@@ -0,0 +1,47 @@
...
+ * A hook for getting the ActiveWorkspace,
+ * and setting the ActiveWorkspace.
+ */
+export const useWorkspaces = () => {
+	const dispatch = useDispatch();
+	const setActiveWorkspace = (activeWorkspace: WorkspaceState['activeWorkspace']) => {
+		dispatch(Actions.setActiveWorkspace(activeWorkspace));
+	}
+	// Run every time the workspace service pushes out an update.
+	const onWorkspaceUpdate = (err: any, response: any) => {
```
d4hines: Because the error isn't used here, it shouldn't be in the function. Also, we have to tighten up these types, but that can be done later.

src-built-in/components/UIAPI/reducers/workspaces.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { assert } from 'chai';
```
d4hines: Do we not like the `expect` or `should` API's?

src-built-in/components/UIAPI/reducers/workspaces.ts
```diff
@@ -0,0 +1,27 @@
...
+				}
+			}
+		default:
+			return state;
+	}
+}
+
+// Wraps the reducer with some logging so that we can debug
+// field reported bugs.
+export default withLogging("Workspaces", workspaces);
```
d4hines: We don't need to do this here. It only needs to be done once at the top level when we create the store.

src-built-in/components/UIAPI/reducers/workspaces.spec.ts
d4hines: This is from Mody. I'll ask him.

src-built-in/components/UIAPI/hooks/workspaces.ts
zona-ciq: If there is an error, we should probably log it out to the console

carterbs: I'll let you guys deal with this later. err is the first argument always for a pub/sub even if you don't use it. Yeah we should log the error if there is one. I think tightening these types will have to wait for typing FSBL. Maybe not though

src-built-in/components/UIAPI/reducers/workspaces.ts
carterbs: what? I'm going to merge this because you guys approved it but please explain.

PR #587, opened 2020-02-06
feat: #id 21653 favorites toolbar component design

zona-ciq:
feat: #id 21653

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/21653/details/)


src-built-in/components/UIAPI/hooks/pins.ts
```diff
@@ -0,0 +1,29 @@
+// import { useEffect, useState } from "react";
```
thorsent: The hook should be "useFavorites", not "usePins". We're deprecating the word "pin" so please refactor that name out.

src-built-in/components/UIAPI/reducers/toolbar.ts
```diff
@@ -0,0 +1,27 @@
+// import { loop, Cmd } from 'redux-loop';
+
+// import { ActionTypes } from '../types';
+// import { loadPinnedItems } from '../effects/toolbar';
+
+// const initialState = {
```
thorsent: I think the state design should be reverse the hierarchy here. Eventually we want to be able to handle *any* type of favorite. So it should be a list of favorites. A favorite itself should have a type which can be {"workspace"|"component"}. This will allow us to add more types in the future.

src-built-in/components/UIAPI/reducers/toolbar.ts
```diff
@@ -0,0 +1,27 @@
...
+//         },
+//         component: {
+
+//         }
+//     }
+// };
+
+// export const toolbar = (state = initialState, action) => {
+//     const { type, payload } = action;
+//     switch (type) {
```
thorsent: The design should include all of the actions necessary for the entire favorites subsystem. Try not to think about one component at a time. Think holistically.

INITIALIZE_FAVORITES
ADD_FAVORITE
REMOVE_FAVORITE
REORDER_FAVORITE
etc

Refer to the requirements document to determine what your actions need to be.

src-built-in/components/toolbarRefactored/favorites.tsx
```diff
@@ -0,0 +1,14 @@
+// import * as React from "react";
+
+// import usePins from "../UIAPI/hooks/pins";
+// import { FinsembleToolbarSection } from "@chartiq/finsemble-react-controls";
+
+// export const Favorites: React.FunctionComponent = () => {
+//     const pinnedItems = usePins("favorites");
+//     return (
+//         // FinsembleToolbarSection needs to be refactored
+//         <FinsembleToolbarSection>
```
thorsent: There shouldn't be a FinsembleToolbarSection. This should just be a div.

Styling will need to be done at the template level and we will no longer be using anything from finsemble-react-controls.

src-built-in/components/UIAPI/reducers/toolbar.ts
zona-ciq: Yes when I was drawing the diagram I realized it should be something like this: 
state

src-built-in/components/UIAPI/hooks/pins.ts
zona-ciq: 👍

src-built-in/components/UIAPI/reducers/toolbar.ts
zona-ciq: Not sure why it's rotated to the left. If you click on the image it will open another page with the right orientation

src-built-in/components/UIAPI/reducers/toolbar.ts
zona-ciq: Got it!

src-built-in/components/toolbarRefactored/favorites.tsx
zona-ciq: Wow this is actually a critical piece of information that I missed. I thought we were going to keep the elements in react control. Thanks for pointing this out. This would help with the entire Favorite design

src-built-in/components/UIAPI/reducers/toolbar.ts
thorsent: You've already taken a step towards a normalized schema, which is great. Probably you're going to need unique identifiers.

```
items : {
   "GSD1212SSD" : {
   "label": "Welcome component",
   "action": "SWITCH_WORKSPACE",
   "payload" : {}
   }
},
order : ["GSD1212SSD","XSD12323A"
]

@d4hines was investigating normalization best practices for redux:
https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape/

zona-ciq: You're right, we'll need IDs for every favorite component. Normalized state should be very useful. I'm aligning my current design to that practice.

PR #588, opened 2020-02-06
18740 linker refactor ian's changes

mesner:
changes:
 - remove actions
 - move effects into hooks file
 - simplify reducer

todo:
 - fix window showing bug
 - fix finsemble provider wrapper (no redux in linker)


src-built-in/components/UIAPI/hooks/linker.ts
```diff
@@ -1,24 +1,92 @@
...
 export const useLinker = () => {
+
     const dispatch = useDispatch();
-    const linker: Linker = useSelector((state: RootState) => state.linker);
+    const state: LinkerState = useSelector((state: RootState) => state.linker);
 
     useEffect(() => {
-		dispatch(actions.init());
+
+        FSBL.Clients.Logger.system.log("Linker component is mounted. Initializing.");
```
carterbs: I prefer this. I thought it was a little strange having an init action.

src-built-in/components/UIAPI/hooks/linker.ts
```diff
@@ -1,24 +1,92 @@
...
+    }, [state])
+
+    const hideWindow = () => {
+        finsembleWindow.hide();
+    };
+    const showWindow = () => {
+        finsembleWindow.focus();
+    };
+
+    const setInitialChannels = () => {
```
carterbs: Would prefer these to be defined before they're called. Just a code organization thing. I may move them if you aren't working today.

src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -1,81 +1,46 @@
...
-        default:
-            return state;
-    }
-}
-
-export default withLogging("Linker", linker);
\ No newline at end of file
+const reducer = (state = initialState, action: ACTION_TYPES) =>
+    produce(state, (draft: LinkerState) => {
+        const draft2 = actions.match(action,{
```
carterbs: Why do we assign a variable (draft2) and never use it?

mesner:
I agree it does many things but I think it is clean.

The hook should ideally encapsulate all of finsemble api for the sake of
simplicity of the consuming component, so while I agree that it is doing
many things in the code of one file, I don't think that separating
different areas of responsibility into different parts will result in
cleaner code.

I'm doing so, you will undoubtedly take 100 lines and turn them into 300
and complicate the control flow. I reject that for this use case. For the
toolbar and more complex components? Maybe, but I like having the fsbl
effects front and center until they can be replaced with remote actions in
a distributed store.

I see the point, am open to suggestions, but am skeptical that there is a
simpler way.

src-built-in/components/UIAPI/hooks/linker.ts
```diff
@@ -1,24 +1,120 @@
...
+            windowIdentifier: msg.data.windowIdentifier,
+        }));
+        msg.sendQueryResponse(null, null);
+    }
+
+    const initAccessibleLinkerMode = () => {
+        const accessibilityCallback = (err: any, value: boolean) => {
+            if (err) FSBL.Clients.Logger.system.error(`Failed to get accessibleLinker value: ${err}`);
+            dispatch(actions.SET_ACCESSIBILITY({
+                isAccessibleLinker: true,
```
zona-ciq: Here it should be set to "value" instead of true. Maybe something like ```value ? value : true```

carterbs: Could do !!value

src-built-in/components/UIAPI/hooks/linker.ts
```diff
@@ -1,24 +1,120 @@
...
+            finsembleWindow.removeEventListener("blurred", hideWindow);
+            finsembleWindow.removeEventListener("shown", focusWindow);
+            FSBL.Clients.RouterClient.removeEventListener("Finsemble.LinkerWindow.SetActiveChannels", onActiveChannelsChanged);
         }
-	}, []);
+    }, []);
+
+    useEffect(() => {
+        FSBL.Clients.WindowClient.fitToDOM();
+    }, [state])
```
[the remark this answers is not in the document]
carterbs: We're going to leave this for now. Will come back to optimize later if needed.

src-built-in/components/UIAPI/hooks/linker.ts
```diff
@@ -1,24 +1,120 @@
...
+
+        // We're activating if the channel is currently inactive
+        const activating = !channelActive;
+        const deactivating = channelActive;
+        const allowedToActivateAnotherChannel = numberOfActiveChannels < MAXIMUM_ALLOWED_ACTIVE_CHANNELS;
+        // if we're toggling a channel to be inactive OR
+        // we're trying to toggle one to be active AND we have not hit our maximum,
+        // let the call go through. Otherwise log an error.
+        if (deactivating) {
+            FSBL.Clients.LinkerClient.unlinkFromChannel(channelName, windowIdentifier, linkCallback(false));
```
zona-ciq: I'd suggest renaming linkerCallback to something like ```updateActivationTo(false/true)```. It's more readable and people would know what that function is doing without looking into it

src-built-in/components/UIAPI/reducers/linker.spec.ts
```diff
@@ -5,162 +5,35 @@
...
+const channel = (id: string, active: boolean, color: string) => {
     return {
         name: color,
         color,
         active,
-        id
+        id,
+        border: color
     }
 }
```
zona-ciq: Are we going to test the linker's reducer anymore?

src-built-in/components/UIAPI/reducers/linker.ts
```diff
@@ -1,81 +1,46 @@
...
+                const channelNameArray = Object.values(channelNames).map(obj => obj.name);
+                draft.windowIdentifier = windowIdentifier;
+                for (const channel of Object.values(draft.channels)) {
+                    channel.active = channelNameArray.includes(channel.name);
+                }
+            },
+            SET_ACCESSIBILITY: ({ isAccessibleLinker }) => {
+                draft.isAccessibleLinker = isAccessibleLinker;
+            },
+            default: a => draft,
```
zona-ciq: What is "a"? This should probably be renamed

src-built-in/components/UIAPI/reducers/workspaces.ts
```diff
@@ -1,27 +1,22 @@
...
-		default:
-			return state;
-	}
+export const reducer = (state = initialState, action: ACTION_TYPES) => {
+	return produce(state, (draft: WorkspaceState) => {
+		actions.match(action, {
+			SET_ACTIVE_WORKSPACE: ({ name }) => {
+				draft.activeWorkspace.name = name;
+			},
+			default: a => draft
```
zona-ciq: Same here

src-built-in/components/UIAPI/reducers/linker.spec.ts
carterbs: Yes ma'am - just not for this PR. It's on the list to go back around. Because it changed so much and we want to get the finsemble-UI up and running, I think it's worth skipping. I will create a card to get the tests rolling.

src-built-in/components/UIAPI/reducers/linker.ts
carterbs: Has been! Good catch

src-built-in/components/UIAPI/reducers/workspaces.ts
carterbs: already done

src-built-in/components/UIAPI/hooks/linker.ts
carterbs: https://github.com/ChartIQ/finsemble-ui/pull/1/commits/4cef90004bdd6de35b8c8274c024e3a31ab076b1

If this is good with you, i'll close out this PR and move the card. Thanks!

carterbs:
Closing PR; work has been moved to finsemble-UI repository. Work has been reviewed.

PR #589, opened 2020-02-06
feat: #id 21748 ancillary requirements

carterbs:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/21748/details)

MUST BE MERGED AFTER THIS PR: https://github.com/ChartIQ/finsemble-seed/pull/586/files

**I branched off of that branch. I have purposefully used that as the base for this PR for review purposes only. Once it's merged, I'll rebase off of seed-refactor so the commit tree doesn't look weird.**

**Description of change**
* Reverted toolbar size.
* Added toolbar hotkeys.
* Reduced the startup footprint of toolbar.
* Publish system checkpoint.


src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
```diff
@@ -0,0 +1,109 @@
...
+	 */
+const bringToolbarToFront = (focus = false) => {
+	finsembleWindow.bringToFront(null, (err:any) => {
+		if (err) {
+			FSBL.Clients.Logger.error(`bringToolbarToFront failed, error:`, err);
+		}
+
+		if (focus) {
+			finsembleWindow.focus();
+			// Leaving this line of code because clearly someone ran into something difficult.
```
zona-ciq: This comment needs to be more specific. It looks like a TODO item for you when the search component is complete. Are we redesigning the search bar? If so does the search bar need to be in focus when we bring up the toolbar using the hotkey?

src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
```diff
@@ -0,0 +1,109 @@
...
+	});
+}
+
+const registerShowSearchHotkey = () => {
+	FSBL.Clients.HotkeyClient.addGlobalHotkey([keys.ctrl, keys.alt, keys.f], (err: any) => {
+		handleHotkeyRegistrationError(err);
+		bringToolbarToFront(true);
+	});
+}
+
```
zona-ciq: Are we keeping the hot key for favorites items? If so we probably need to set up the favorite hotkeys here as well.

src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
carterbs: We aren't redesigning the search as far as I know. I have no idea what this line does but it seemed important. what exactly are you looking for in the comment?

src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
carterbs: So, they are not in the requirements, so I didn't add them. They're a big piece of complication. I can ask terry and Sean if you'd like.

src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
zona-ciq: This line refocuses the search bar whenever someone uses the hotkey to bring up the toolbar. I'm saying if we want to keep this functionality we should refactor it when the search bar and the toolbar state management is setup.

src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
zona-ciq: Sure! In case we missed a functionality when writing the requirement

src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
carterbs: setActive false is refocusing the search bar? My comment was about the line of code below.

src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
carterbs: Okay, discussing with them. I think it's a super undiscoverable feature and should be removed in either case. I didn't know anyone outside of me and nicolai knew it existed.

src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
carterbs: @zona-ciq updated the comment. please take a look

zona-ciq: Resolved on the phone

src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
```diff
@@ -0,0 +1,109 @@
...
+		}
+	});
+}
+
+ 	/**
+* Hides the toolbar
+*/
+const hideToolbar = () => {
+ finsembleWindow.blur();
+ finsembleWindow.hide();
```
d4hines: Shouldn't this be in the callback to `blur`? I know they're on a queue internally, but we shouldn't presume that knowledge here.

src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
```diff
@@ -0,0 +1,109 @@
...
+ */
+const revealAllWindows = () => {
+	FSBL.Clients.LauncherClient.bringWindowsToFront()
+}
+
+const registerShowToolbarHotkey = () => {
+	FSBL.Clients.HotkeyClient.addGlobalHotkey(
+		[keys.ctrl, keys.alt, keys.t],
+		(err: any) => {
+			handleHotkeyRegistrationError(err);
```
d4hines: Is this callback called on hotkey registration or every tiem the key is pressed? If the latter, isn't `handleHotkeyRegistrationError` a misnomer?

src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
carterbs: No sir; we want it to hide immediately. We blur just so that any open menus will also.

d4hines: I'm a dumb-dumb.

src-built-in/components/toolbarRefactored/toolbarHotkeys.ts
d4hines: Answer: this is just how this works.

PR #602, opened 2020-02-20
Window titlebar should not register with docking, since the window re…

clausjs:
fix: #id [22315]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/22315/details)

**Description of change**
* Window title bar should not register with docking since the window registers itself with the windowClient, removed the call to register the titlebar with empty options

**Description of testing**
1. Disable snapping in a component's config
1. [ ] Ensure snapping actually disabled

### DIFF
--- a/src-built-in/components/windowTitleBar/src/windowTitleBarComponent.jsx
+++ b/src-built-in/components/windowTitleBar/src/windowTitleBarComponent.jsx
@@ -454,8 +454,6 @@ function init () {
 		HeaderActions = storeExports.Actions;
 		windowTitleBarStore = storeExports.getStore();
 		ReactDOM.render(<WindowTitleBar />, FSBLHeader);
-		// Register with docking manager
-		FSBL.Clients.WindowClient.registerWithDockingManager();
 	});
 }
 


Garbee:
Looks like this line was introduced with [allowing the preload](https://github.com/ChartIQ/finsemble-seed/commit/9f60e36b2d5ed21909b6b7039723c6c2a934135d) of the titlebar. Was this tested with the various methods (inject, preload, any others?) to make sure no adverse affects are incurred with windows between the methods?

clausjs:
> Looks like this line was introduced with [allowing the preload](https://github.com/ChartIQ/finsemble-seed/commit/9f60e36b2d5ed21909b6b7039723c6c2a934135d) of the titlebar. Was this tested with the various methods (inject, preload, any others?) to make sure no adverse affects are incurred with windows between the methods?

I wouldn't even know how to test this, I do know this is an issue I've fixed before because the same thing was happening. A component registers twice, the second time (from the titlebar) passes in bad information which overwrites the _correct_ configuration which was passed in on the first time.
Let me do a little research

clausjs:
It looks like most of the changes you reference were to address an issue with the Floating Titlebar (which we don't have in electron). If we have to re-implement it one day, we'll need to find another way that doesn't require re-registering. If this is the only way, then we need to make sure the titlebar registration also passes in the component's config which has already been registered at this point.

PR #603, opened 2020-02-21
feat: #id 21598 refactor RevealAll, MinimizeAll and AutoArrange

meredith-ciq:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/21598/details)

Requires: https://github.com/ChartIQ/finsemble-ui/pull/6

**Description of change**
* Paired with pr to finsemble-ui to wrap these components
* Adds the updated components to toolbar shell.
* Temporary solution for adding toolbar sections - will need to work with team to find a longterm solution.

### DIFF
--- a/src-built-in/components/toolbar/src/app.jsx
+++ b/src-built-in/components/toolbar/src/app.jsx
@@ -4,11 +4,7 @@
 */
 import ReactDOM from "react-dom";
 import React from "react";
-import { ToolbarShell } from "@chartiq/finsemble-ui/lib/components/toolbar/ToolbarShell";
-import { RevealAll } from '@chartiq/finsemble-ui/lib/components/toolbar/RevealAll';
-import { MinimizeAll } from '@chartiq/finsemble-ui/lib/components/toolbar/MinimizeAll';
-import {AutoArrange } from '@chartiq/finsemble-ui/lib/components/toolbar/AutoArrange';
-import { ToolbarShell, DragHandle } from "@chartiq/finsemble-ui/lib/components";
+import { ToolbarShell, DragHandle, RevealAll, MinimizeAll, AutoArrange } from "@chartiq/finsemble-ui/lib/components";
 
 import '../toolbar.css';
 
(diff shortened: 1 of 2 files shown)


src-built-in/components/toolbar/src/app.jsx
```diff
@@ -12,20 +12,27 @@ import React from "react";
...
-			<DragHandle/>
-			<DummyExport/>
-			{/* Drag Handle */}
-			{/* Workspace Management Menu */}
-			{/* App Menu */}
-			{/* Favorites Section */}
-			{/* Minimize All */}
-			{/* Reveal All */}
-			{/* Auto Arrange */}
+			<div className="finsemble-toolbar-section left">
```
d4hines: Should we do `div`'s like this or have a specific component like
```jsx
import { ToolbarSection } from "@chartiq/finsemble-ui/components
...
<ToolbarShell>
   <ToolbarSection position=left>
    ...
   </ToolbarSection>
   <ToolbarSection position=center>
    ...
   </ToolbarSection>
   <ToolbarSection position=right>
    ...
   </ToolbarSection>
</ToolbarShell>
```

I can see pro's and con's to both.

zona-ciq: I like just doing the div - it gives our client developers more options to customize

src-built-in/components/toolbar/src/app.jsx
```diff
@@ -12,20 +12,27 @@ import React from "react";
...
+				{/* Workspace Management Menu */}
+				{/* App Menu */}
+			</div>
+			<div className="finsemble-toolbar-section center">
+				{/* Favorites Section */}
+			</div>
+			<div className="finsemble-toolbar-section right">
+				<RevealAll />
+				{/* Minimize All */}
+				{/* Auto Arrange */}
```
zona-ciq: If we're disabling those two functionalities by commenting this code out, should we import them at the top of this file so people can just use them when those lines are uncommented?

src-built-in/components/toolbar/src/app.jsx
meredith-ciq: This was meant to be a shortterm solution, we need to decide as a team which we want.

d4hines: Terry decided Meredith's current way is best.

PR #607, opened 2020-02-27
feat: #id 21586 search

carterbs:
Rewrote search to match this: https://app.zeplin.io/project/5a6b7dea31e01cd724fc9f60/screen/5e470adc0a57c3aa55352987

KB: https://chartiq.kanbanize.com/ctrl_board/18/cards/21586/details/


configs/application/manifest-local.json
```diff
@@ -1,4 +1,14 @@
 {
+    "finsemble-electron-adapter": {
```
d4hines: I don't think this is upposed to be here.

PR #617, opened 2020-03-05
feat: #id 22724 wrap advancedAppCatalog

amaroks:
https://chartiq.kanbanize.com/ctrl_board/18/cards/22724/details/

Testing:
1. Checkout finsemble-ui@`22724-wrap-advancedcatalog`
1. Checkout seed@`22724-wrap-advancedcatalog`
1. Edit seed's config.json and set `appDirectoryEndpoint` to http://localhost:3030
1. clone and run https://github.com/ChartIQ/fdc-appd
1. In toolbar's devtools run the following command to spawn app catalog
```javascript
FSBL.Clients.LauncherClient.spawn("Advanced App Catalog")
```

You can still test this without 3 and 4 but you will only see a message telling you about an error retrieving data.


amaroks:
@d4hines I have renamed AppMarket to AppCatalog. Unfortunetally waitForFSBL does not work for my case (when component is passed to `ReactDOM.render` and fixing that isn't part of this card so we can revisit later.

PR #620, opened 2020-03-06
refactor: #id 22721 Authentication component seed refactor

zona-ciq:
refactor: #id 22721

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/22721/details/)

**Description of change**
* Take out the original authentication component and make the new one using a template form that enables better customization and internal control.

**Description of testing**
1. Enables authentication for finsemble. Refer to this [doc](https://documentation.chartiq.com/finsemble/tutorial-Authentication.html)for how to do the setup.
1. Test this branch along with branch 22721-authentication-refactor for finsemble-ui repo
1. Launch finsemble. Make sure the authentication window comes up first, and upon submit the form closes and the rest of the application launches.


src-built-in/components/authentication/app.jsx
```diff
@@ -0,0 +1,93 @@
...
+}
+
+
+function renderAuthComponent() {
+    ReactDOM.render(<Authentication />, document.getElementById("authentication"));
+    FSBL.System.Window.getCurrent().show();
+}
+
+if (window.FSBL && FSBL.addEventListener) {
+	FSBL.addEventListener("onReady", renderAuthComponent);
```
zona-ciq: Since we're using the Authentication component written in this file instead of importing it from finsemble-ui, there's no better way to hide the FSBL onReady function from this file.

d4hines: We could wrap AuthForm and AuthHeader in WaitForFSBL. I suggest we do this.

src-built-in/components/authentication/app.jsx
```diff
@@ -0,0 +1,93 @@
+/*!
+* The authentication component is a React component that renders a user form to collect auth info and determine whether to give user access to the rest of finsemble
```
d4hines: Use caps for consistency.
```suggestion
* The authentication component is a React component that renders a user form to collect auth info and determine whether to give user access to the rest of Finsemble
```

src-built-in/components/authentication/app.jsx
```diff
@@ -0,0 +1,93 @@
...
+                <div className="fsbl-login-logo" alt="Finsemble Sign In"></div>
+                <div className="fsbl-button-wrapper">
+                    {error && <ErrorMessage>
+                        {error}
+                    </ErrorMessage>}
+                    <AuthInput
+                        type="text"
+                        name="username"
+                        placeholder="Username"
+                        onChange={handleChange}
```
d4hines: What kind of customizations do we expect users to make to this function?

Do they need to cancel/modify input? If not, then we should call `handleChange` inside `AuthInput` and make `onChange` optional, instead of requiring they always pass the same function in.

d4hines: I'm sorry I didn't bring this up in the design - I didn't get a chance to look at the doc.

zona-ciq: We need to capture the form input change and store them inside formValues for sending the information to the auth server. This is to make it a React controlled form input element. I've changed the function name to make its purpose clearer.

src-built-in/components/authentication/app.jsx
```diff
@@ -0,0 +1,84 @@
...
+            setError(authResult.error);
+        }
+    };
+
+    return (
+        <>
+            <AuthHeader>
+                <img className="fsbl-logo" />
+                <div className="fsbl-close">
+                    <i className="ff-close" onClick={quitApplication}></i>
```
carterbs: Please try this out. I THINK the onClick should be on the parent div. I shouldn't have to click on the lines of the X in order to close the authentication component. I should be able to click on the padding around the x. If this doesn't make sense let me know.

src-built-in/components/authentication/app.jsx
```diff
@@ -0,0 +1,84 @@
...
+                </div>
+            </AuthHeader>
+
+            <AuthForm>
+                <div className="fsbl-login-logo" alt="Finsemble Sign In"></div>
+                <div className="fsbl-button-wrapper">
+                    {error && <ErrorMessage>
+                        {error}
+                    </ErrorMessage>}
+                    <AuthInput
```
carterbs: I think the auth template should look very small.

```
<AuthUserName/>
<AuthPassword/>
```

What's the justification for these AuthInput components?

src-built-in/components/authentication/app.jsx
zona-ciq: It doesn't matter. If you put the onclick handler on the icon the padding area will still detect onclick event. I think you were thinking of an `<svg>` situation

src-built-in/components/authentication/app.jsx
zona-ciq: In case the developers want to add other forms of input, such as lastname, firstname, we will be able to handle it without changing our code. If we're being too specific about the React component we provides, we will need to write additional `<AuthFirstname />` or `<AuthLastname />`, etc.

zona-ciq: In case the developers want to add other forms of input, such as lastname, firstname, we will be able to handle it without changing our code. If we're being too specific about the React component we provides, we will need to write additional `<AuthFirstname />` or `<AuthLastname />`, etc. This is a typical way of providing React component templates (see Material.UI)

PR #625, opened 2020-03-12
fix: Seed-refactor, resolve build issues

carterbs:
* Resolves issues with hardSourceLoader causing `undefinedResult` if you didn't start with a fresh cache. Caching is built into webpack now, so we can remove it. Left `dev-fresh` in case we need to delete webpack's cache.
* Adds source-map-loader as suggested by Typescript's official docs. Breakpoints now work!: https://www.typescriptlang.org/docs/handbook/react-&-webpack.html
* Removes ^s in some packages.


carterbs:
@zona-ciq fixed! What an amateur move.

I did not look at the performance. I was mainly trying to remove the hurdle of having to hit `dev-fresh` every build. I will create a card to assess build performance from 4.1 to `seed-refactor` and try to get it into the next sprint. let me know if that works.

PR #626, opened 2020-03-12
fix: #id 22271 Take out depreciated toolbar config

zona-ciq:
fix: #id 22271

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/22271/details/)

**Description of change**
* Took out depreciated toolbar config
* Set the "autoShow" flag inside the toolbar config to be true in case user set the workspaceService.global to be false (currently undocumented). In that case if we have autoShow set to be false, the toolbar would come up hidden.

**Description of testing**
1. Launch finsemble
1. Make sure toolbar shows up and cannot tab, tile, nor snap

### DIFF
--- a/configs/application/UIComponents.json
+++ b/configs/application/UIComponents.json
@@ -595,7 +595,7 @@
 				"docked": "top",
 				"canGroup": false,
 				"options": {
-					"autoShow": false,
+					"autoShow": true,
 					"contextMenu": false,
 					"showTaskbarIcon": false,
 					"smallWindow": true,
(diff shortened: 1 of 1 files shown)


configs/application/UIComponents.json
```diff
@@ -595,7 +595,7 @@
 				"docked": "top",
 				"canGroup": false,
 				"options": {
-					"autoShow": false,
+					"autoShow": true,
```
carterbs: @zona-ciq why did you change this again? It should probably be in the PR description.

carterbs:
@zona-ciq how does changing autoshow impact startup appearance? IIRC we had the toolbar hidden because it comes up, figures out where it needs to go, and then moves itself. That, and we didn't want a white toolbar on startup.

configs/application/UIComponents.json
zona-ciq: I just tested with workspaceService.global set to true and toolbar did come up initially blank. I'll change it back for now. When we document the `global` attribute we need to be aware that if the `autoShow` field is set to "false" the toolbar wouldn't automatically show up on startup.

PR #628, opened 2020-03-16
feat: #id 22263 Make the enabling of app catalog configurable

zona-ciq:
feat: #id 22263

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/22263/details/)

**Description of change**
* Add in the "enabled" field for the UIComponents.App Catalog config entry
* Change the app catalog src-built-in code so that if the "enabled" field is set to be "true", the App Catalog will appear in the Advanced App Launcher

**Description of testing**
1. Enable the advanced app launcher
1. Go to finsemble seed -> config -> application -> UIComponents.json, under the "advanced app launcher" -> "component" field, add `"useAppCatalog": true`.
1. Launch finsemble and notice that the App Catalog will appear in the Advanced App Launcher

### DIFF
--- a/src-built-in/components/advancedAppLauncher/src/components/LeftNavBottomLinks.jsx
+++ b/src-built-in/components/advancedAppLauncher/src/components/LeftNavBottomLinks.jsx
@@ -1,11 +1,10 @@
 import React from 'react'
 import { getStore } from '../stores/LauncherStore'
-
+import * as UIconfig from '../../../../../configs/application/UIComponents.json'
 
 const bottomEntries = [
 	{ name: "New App", icon: "ff-new-workspace", click: "showAddAppForm" },
 	//{ name: 'New Dashboard', icon: 'ff-dashboard-new' },
-	// { name: "App Catalog", icon: "ff-list", click: "openAppMarket" }
 ];
 
 export default class LeftNavBottomLinks extends React.Component {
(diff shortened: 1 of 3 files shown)


src-built-in/components/advancedAppLauncher/src/components/LeftNavBottomLinks.jsx
```diff
@@ -1,18 +1,38 @@
...
+				//{ name: 'New Dashboard', icon: 'ff-dashboard-new' },
+			]
+		}
+	}
+
+	componentWillMount() {
+		// If user configures the advanced app launcher to use the app catalog, add app catalog button to the bottom entries of the
+		// advanced app launcher menu.
+		finsembleWindow.getOptions((_, opts) => {
+			const useAppCatalog = (((opts || {}).customData || {}).component || {}).useAppCatalog;
```
amaroks: A very confusing and hard to read line. What is the criteria you're looking for?

zona-ciq: Here I'm using Oliver Steele’s Nested Object Access Pattern. Basically when reading such lines you just ignore the `{}`, `||`, and `()` and just ready it like `opts.customData.component.useAppCatalog`. I didn't use lodash.get here because I don't want to hid the object's path. If you don't think we need to show it we can change to using lodash.get

zona-ciq:
> If we need a place to turn on config in the seed file, we probably should put this logic somewhere else.

I'm not sure what changes are you requesting. The config does live in the seed.

meredith-ciq:
Yes, you're right that we don't more config. This file has been moved to finsemble-ui and should be deleted soon, so there's no need to make the changes in seed and the ui.

PR #632, opened 2020-03-17
22246 app catalog ui updates

jason-ciq:
style: #id [22246]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/22246/details)

**Changed Remove from trash can to button**

**Description of testing**

1. Run Finsemble
1. Open Advanced App Catalog and add an App
1. On the detail page, verify that we have a "Remove" button where the Trash can was, and that it's styled correctly.

### DIFF
--- a/src-built-in/components/advancedAppCatalog/src/components/Showcase/Header.jsx
+++ b/src-built-in/components/advancedAppCatalog/src/components/Showcase/Header.jsx
@@ -69,7 +69,9 @@ const Header = props => {
 						)}
 				</button>
 				{props.installed && <button className={props.entitled ? "action-button remove" : "action-button remove disabled"} disabled={!props.entitled} onClick={removeApp}>
-					<div className='ff-adp-trash-outline' title='Remove App'/>
+					<div title='Remove App'>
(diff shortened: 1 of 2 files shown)


src-built-in/components/advancedAppCatalog/src/components/Showcase/Header.jsx
```diff
@@ -69,7 +69,9 @@ const Header = props => {
 						)}
 				</button>
 				{props.installed && <button className={props.entitled ? "action-button remove" : "action-button remove disabled"} disabled={!props.entitled} onClick={removeApp}>
-					<div className='ff-adp-trash-outline' title='Remove App'/>
+					<div title='Remove App'>
```
Garbee: Pre-existing issue, this isn't something to solve here. However this is bad accessibility.

1) This should just be a `<button>`
2) Title shouldn't be necessary when the textContent is clear or the element specifically kinda needs it for the a11y tree (`<nav>` for example.)

clausjs: It is a button

PR #637, opened 2020-03-23
Remove OF from build, comments, package.json, etc

carterbs:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/23020/details)

* See card for details


configs/application/config.json
```diff
@@ -58,7 +61,7 @@
 			"unknownComponent": "nonConfiguredComponent"
 		},
 		"assimilation": {
-			"useOpenFinSpawn": false,
+			"useFeaSpawn": false,
```
zona-ciq: In assimilation window manager we're still using useOpenFinSpawn. This will cause that code to break

carterbs: Not in 5.0.0

carterbs: https://github.com/ChartIQ/finsemble/commit/c1f36d12d96aeaac29b9854d9c0ef04bbfdb44df

zona-ciq: Cool!

PR #641, opened 2020-03-25
22360 Refactor Titlebar

d4hines:
refactor: #id [CARD]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/22360/details)

**Description of change**
Moved titlebar components and logic to finsemble-ui.

Special items of note:
- I rewrote AlwaysOnTopButton.
- Users have to pass in their Titlebar into a special function `initializeTitlebar`, which does the injection of the DOM. The alternative is to keep the injection code (5 lines) in the template file. I went with this route because it keeps the seed as minimal as possible; however, it has the drawback that we have to have this special init function, which none of the other components have. If you feel I made the wrong choice, let's talk about - I don't feel strongly one way or the other.

**Description of testing**
- I enabled every button on the titlebar via config, then disabled one button at a time and tested they disabled/enabled correctly. I did this for Linker, Sharer, AlwaysOnTop, Docking, and Minimize.
- I verified each requirement from this doc was met: https://docs.google.com/document/d/1ivIJotCybdArSm29Mo2MtvJvAv4245XsLEYKebrVWls/edit


build/webpack/defaultWebpackConfig.js
```diff
@@ -26,7 +26,7 @@ module.exports = class WebpackDefaults {
 
 		if (env === "production") {
 			// When building the production environment, minify the code.
-			plugins.push(new UglifyJsPlugin());
+			// plugins.push(new UglifyJsPlugin());
```
carterbs: is the env set incorrectly?

build/webpack/webpack.titleBar.js
```diff
@@ -11,17 +10,14 @@ let plugins =
...
 
 if (env === "production") {
     // When building the production environment, minify the code.
     plugins.push(new UglifyJsPlugin());
 }
 
-var builtInTitleBarPath = "./src-built-in/components/windowTitleBar/src/windowTitleBar.jsx";
-var customTitleBarPath = "./src/components/windowTitleBar/src/windowTitleBar.jsx";
-var titleBarPath = fs.existsSync(customTitleBarPath) ? customTitleBarPath : builtInTitleBarPath;
+var titleBarPath  = "./src-built-in/components/titlebar/titlebarLoader.js";
```
carterbs: @zona-ciq make sure this gets changed in your PR

src-built-in/components/titleBar/TitleBar.jsx
```diff
@@ -0,0 +1,54 @@
...
+				{showAlwaysOnTopButton && <AlwaysOnTopButton/>}
+				{showMinimizeButton && <MinimizeButton />}
+				{showMaximizeButton && <MaximizeButton />}
+				{showCloseButton && <CloseButton />}
+			</div>
+		</>
+	);
+}
+
+initializeTitleBar(TitleBar);
```
carterbs: can you remind me why this initialization logic couldn't be wrapped in TItleBarShell?

d4hines: The reason is that it injects the DOM to which the titlebar is mounted. Without it, where would we mount? The conclusion we reached over call was to rename to `initializeTitlebar` to `mountTitlebar`.

PR #648, opened 2020-04-01
refactor: #22360 Refactor Title Bar

d4hines:
refactor: #id [22360]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/22360/details)

**Description of change**
Moved titlebar components and logic to finsemble-ui.

Special items of note:
- I rewrote AlwaysOnTopButton.
- Users have to pass in their Titlebar into a special function `initializeTitlebar`, which does the injection of the DOM. The alternative is to keep the injection code (5 lines) in the template file. I went with this route because it keeps the seed as minimal as possible; however, it has the drawback that we have to have this special init function, which none of the other components have. If you feel I made the wrong choice, let's talk about - I don't feel strongly one way or the other.

**Description of testing**
- I enabled every button on the titlebar via config, then disabled one button at a time and tested they disabled/enabled correctly. I did this for Linker, Sharer, AlwaysOnTop, Docking, and Minimize.
- I verified each requirement from this doc was met: https://docs.google.com/document/d/1ivIJotCybdArSm29Mo2MtvJvAv4245XsLEYKebrVWls/edit


configs/application/config.json
```diff
@@ -80,10 +80,10 @@
 			"undockDisbandsEntireGroup": false,
 			"fillHolesOnUndock": true,
 			"tabbing": {
-				"enabled": false
+				"enabled": true
```
zona-ciq, line 83: Tabbing and tiling should be disabled by default. Please double check with Sean to make sure

configs/application/config.json
```diff
@@ -80,10 +80,10 @@
...
 			},
 			"tiling": {
-				"enabled": false
+				"enabled": true
```
zona-ciq, line 86: So is this one.

configs/application/config.json
d4hines: Fixed.

configs/application/config.json
d4hines: Fixed.

PR #651, opened 2020-04-06
fix: #id 22626 Hover colors on active check need color differentiation

clausjs:
fix: #id [22626]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/22626/details)

**Description of change**
* Adds another css class for when a checkmark is already active and gets hovered
* Added code to the AppCard to change the css state of the check depending on the state of installation and hover

**Description of testing**
1. Setup Finsemble and the advanced app launcher/catalog
1. Install an app
1. From a catalog screen containing app cards, notice the active checkmark (green to indicate installed). Hover over it and notice the change in green (subtle difference between light green and a darker green)
1. [ ] Hover state of active checkmark changed on hover

### DIFF
--- a/src-built-in/components/advancedAppCatalog/src/components/AppCard.jsx
+++ b/src-built-in/components/advancedAppCatalog/src/components/AppCard.jsx
@@ -157,7 +157,8 @@ class AppCard extends Component {
 		let { appName, checkShown, checkHighlighted } = this.state;
 
 		let imageIconClasses = "ff-check-mark-2";
-		if (this.props.installed || checkHighlighted) imageIconClasses += " highlighted"
+		if (this.props.installed && checkHighlighted) imageIconClasses += " highlighted added";
+		else if (this.props.installed || checkHighlighted) imageIconClasses += " highlighted"
 		else imageIconClasses += " faded";
 
 		let titleClass = this.state.titleUnderlined ? "app-title highlighted" : "app-title";
--- a/src-built-in/components/advancedAppCatalog/appCatalog.css
+++ b/src-built-in/components/advancedAppCatalog/appCatalog.css
@@ -269,6 +269,11 @@
 	color: var(--catalog-card-checkmark-selected-color);
 }
 
+.app-card .ff-check-mark-2.highlighted.added {
+	display: block;
+	color: var(--accent-positive-1);
+}
+
 .app-catalog {
 	display: flex;
 	flex-direction: column;


src-built-in/components/advancedAppCatalog/src/components/AppCard.jsx
```diff
@@ -157,7 +157,8 @@ class AppCard extends Component {
 		let { appName, checkShown, checkHighlighted } = this.state;
 
 		let imageIconClasses = "ff-check-mark-2";
-		if (this.props.installed || checkHighlighted) imageIconClasses += " highlighted"
+		if (this.props.installed && checkHighlighted) imageIconClasses += " highlighted added";
+		else if (this.props.installed || checkHighlighted) imageIconClasses += " highlighted"
```
stephen-ciq, line 161: It seems weird to me that we have one class for when app is installed and the check is highlighted and one class for when only one of those is true.

I would think that we would want three classes: one for when app is installed, one for when the check is highlighted, and one for when both are true.  What do you think?  I am asking because reading this code just seems counterintuitive to me; I realize that this is a requirements question.  I read [the card ](https://chartiq.kanbanize.com/ctrl_board/18/cards/22626/details/) and it didn't seem clear to me whether we wanted that.  I don't have an account with Zeplin, though, so maybe those links have more clarifying info...

clausjs: So basically there's 3 states that require the css.

- Check is showing for an uninstalled app (gray checkmark), when hovering over a gray checkmark you'll see the green checkmark
- Check is showing for an installed app (green checkmark)
- Check is showing for an installed app and hovered upon (slightly different green checkmark).

In other words, you'll only see `highlighted added` when hovering over a checkmark of an installed app. `highlighted` will apply anytime you hover regardless of the installation status of an app

stephen-ciq: Why is it that we use the same checkmark for when you hover over an uninstalled app and when an app is installed?  Is it because clicking the checkmark installs it?

clausjs: Yes, correct

DanNicolai:
It works!

src-built-in/components/advancedAppCatalog/src/components/AppCard.jsx
DanNicolai: @stephen-ciq the code as it is seems to fulfill what I was looking for. You'll see in the zeplin document that I was going for this set of circumstances; 
uninstalled & no hover: core-primary-4 (gray)
uninstalled & hover: accent-primary
installed & nohover: accent primary
installed & hover: accent-primary-1

htonus:
LGTM

PR #659, opened 2020-04-16
fix: Fix bug with redux complaining about NODE_ENV being set improperly

carterbs:
…ebar (thats where the error was)

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/15542/details)

**Description of change**
* Makes webpack builds consistently use define plugin instead of a mix of define and environment plugin.
* Adds mode to titlebar (that's where the error was).

**Description of testing**
1. Ran seed. Opened a couple components. No errors.

### DIFF
--- a/build/webpack/webpack.titleBar.js
+++ b/build/webpack/webpack.titleBar.js
@@ -96,6 +96,7 @@ module.exports = {
             },
         ]
     },
+    mode: env,
     plugins: plugins,
     output: {
         filename: "[name].js",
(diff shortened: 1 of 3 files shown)


build/webpack/webpack.titleBar.js
```diff
@@ -96,6 +96,7 @@ module.exports = {
             },
         ]
     },
+    mode: env,
```
carterbs, line 99: this fixed the bug

PR #661, opened 2020-04-17
feat: #id 20839 window open works with finsemble

carterbs:

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/20839/details)

**Description of change**
* Config Changes
* Adds portals POC into the seed as an example.

**Description of testing**
1. See core.


src/components/portals/README.md
```diff
@@ -0,0 +1,100 @@
...
+childFinsembleWindow.focus();
+```
+
+### Retrieving the bounds of a child window
+
+Retrieving the bounds of a child window can be done in a couple of ways.
+
+In the browser, you can access `childWindow.screenX` and `childWindow.screenY` (NOTE: These are childWindows, not childFinsembleWindows).
+
+Because of a bug in Electron, these values are off due to scaling issues. Instead, you should use the `finsembleWindow`.
```
mesner, line 61: i) I assume this won't be merged, so this isn't a big deal, but...
a) when we say a bug exists, we should link to it
b) I don't think window.screenX and window.screenY have a bug in chromium or electron.  The windows operating system handles scaling with a scaling factor, and that can lead to unexpected values.  Please let me know if I'm wrong here.

carterbs:
> Where will the portals example go?

I think they should be merged into the seed. hence this PR. One great way to end up with stale code that doesn't work is to keep it out of the seed. I think this should be part of our standard offering, right next to the welcome component.

src/components/portals/README.md
carterbs, line 61:
```suggestion
Because of a problem rationalizing scaled values, screenX and screenY will not line up with finsemble's internal bounds. Instead, you should use the `finsembleWindow`.
```

carterbs: Changed wording.

carterbs:
I've changed my mind about portals example. It adds to the clutter of the seed, which is counter to the primary goal of the refactor. I'll move it back to recipes/tearouts.

PR #662, opened 2020-04-17
fix: #id 24413 fix null reference error in appLauncher

d4hines:
fix: #id [22413]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/24413/details)

**Description of change**

Added a check for null to `customData.spawnData` in `filterComponents` function.

**Description of testing**

I booted up Finsemble and verified the error wasn't there.


thorsent:
This was manually fixed by converting to `FSBL.Clients.WindowClient.getSpawnData()`

PR #668, opened 2020-05-11
Terry cleanup

thorsent:
Terry's cleanup.


src/components/toolbar/src/Toolbar.jsx
```diff
@@ -27,7 +27,6 @@ const Toolbar = () => {
 			<ToolbarSection className="right">
 				<div className="divider"></div>
 				<MinimizeAll />
-				<AutoArrange />
```
mesner, line 30: I don't understand.  Why are we removing this?  Also, it is still shown here https://github.com/ChartIQ/finsemble/blob/terry-pass-at-ui-docs/docs/documents-source/tutorials/Finsemble-UI.md#L46

mesner: Also, the autoarrange component still exists, and just renders an empty div.  All of this seems like it could be improved.
https://github.com/ChartIQ/finsemble-ui/blob/terry-cleanup/src/components/toolbar/AutoArrange.tsx#L33

PR #673, opened 2020-05-15
feat: build improvements

carterbs:
**Description of change**
* Adds HardSourceWebpackPlugin to enable caching between builds
* Makes babel cache properly.
* Makes webpack run in parallel.

**Drops initial build times from 22 s to 9s. Subsequent builds from 22s to 4s.**

**Description of testing**
1. Pulled up finsemble..
1. Made sure watch was working.
1. Made sure build terminated in `npm run build`.
1. Made sure dev-fresh worked.
1. Made sure source maps were workable.


build/buildHelpers.js
```diff
@@ -0,0 +1,202 @@
+const webpack = require("webpack");
```
carterbs, line 1: Everything except the parallel stuff at the bottom was just copy/pasted from gulpfile in an effort to slim that file down.

build/buildHelpers.js
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * Given an array of webpack parallel config objects, this function will spool up a maximum
+ * of N workers (where N is limited by CPU cores). When all of them finish building, the done
+ * callback will be invoked.
+ * 
+ * @param {WebpackParallelConfig[]} WebpackParallelConfigs An array of objects that describe a webpack config
+ * @param {Function} done 
+ */
+const runWebpackInParrallel = (WebpackParallelConfigs, exitOnCompletion, done) => {
```
d4hines, line 178:
```suggestion
const runWebpackInParrallel = (webpackParallelConfigs, exitOnCompletion, done) => {
```
The UpperCamelCase makes me think it's some sort of type or class or something.

gulpfile.js
```diff
@@ -526,7 +370,7 @@ const { launch, connect } = require('hadouken-js-adapter');
 			done();
 		},
 		launchApplication: done => {
-			if (!isNodeVersionValid()) {
+			if (!isNodeVersionValid(MAX_NODE_VERSION)) {
```
d4hines, line 373: Good catch.

build/buildHelpers.js
carterbs: Mistake on my part using ctrl+d. good catch

gulpfile.js
carterbs: actually not a catch. previous MAX_NODE_VERSION was a global defined in gulpfile. I moved the isNodeVersionValid function out of gulpfile.

PR #678, opened 2020-05-20
feat: #id 25975 additional typescript support

WatsonCIQ:
[Kanban link](https://chartiq.kanbanize.com/ctrl_board/106/cards/25995/details/)

**Description of change**
Added the ability to build typescript files when using preload and services from src
* Changes
- Webpack configs changed and tsconfig added react for tsx
- updated typscript dependency to latest (3.9.3)

**Description of testing**
- Create a typescript service, it should build correctly in the dist folder.
- Create a typescript preload, add it to the preloads.entries file and it should build correctly to the dist folder. 
- Try the above with JSX - preloads should build

### DIFF
--- a/build/webpack/webpack.services.js
+++ b/build/webpack/webpack.services.js
@@ -2,7 +2,7 @@ const path = require('path');
 const fs = require("fs");
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const glob_entries = require('webpack-glob-entries');
-const services = glob_entries(path.join(__dirname, '../../', "/src/services/**/*.js"));
+const services = glob_entries(path.join(__dirname, '../../', "/src/services/**/*.*s"));
 let entry = {};
 
 //process service files found
--- a/build/webpack/webpack.services.js
+++ b/build/webpack/webpack.services.js
@@ -2,7 +2,7 @@ const path = require('path');
 const fs = require("fs");
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const glob_entries = require('webpack-glob-entries');
-const services = glob_entries(path.join(__dirname, '../../', "/src/services/**/*.*s"));
+const services = glob_entries(path.join(__dirname, '../../', "/src/services/**/*.{js,ts}"));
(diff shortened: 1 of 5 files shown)


build/webpack/webpack.services.js
```diff
@@ -2,7 +2,7 @@ const path = require('path');
 const fs = require("fs");
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const glob_entries = require('webpack-glob-entries');
-const services = glob_entries(path.join(__dirname, '../../', "/src/services/**/*.js"));
+const services = glob_entries(path.join(__dirname, '../../', "/src/services/**/*.*s"));
```
WatsonCIQ, line 5:
```suggestion
const services = glob_entries(path.join(__dirname, '../../', "/src/services/**/*.{js,ts}"));
```

thorsent:
@sgd2z I've manually implemented these build changes into 5.0 in case you want to just skip the 4.x line.

@carterbs (or @WatsonCIQ ) is there any reason why `webpack.preloads.js` can't use the config from `defaultWebpackConfig.js`? (they look very close)

WatsonCIQ:
> @sgd2z I've manually implemented these build changes into 5.0 in case you want to just skip the 4.x line.
I would say it's worth keeping for 4.x for clients who may take a little longer to upgrade to 5.0 ( I think there are two so far that have already added this fix to their seed project). 

> @carterbs (or @WatsonCIQ ) is there any reason why `webpack.preloads.js` can't use the config from `defaultWebpackConfig.js`? (they look very close)
I think it should be fine to do so, I didn't want to change the code too much for this PR but I don't see why not!

sgd2z:
> @sgd2z I've manually implemented these build changes into 5.0 in case you want to just skip the 4.x line.
> 
> @carterbs (or @WatsonCIQ ) is there any reason why `webpack.preloads.js` can't use the config from `defaultWebpackConfig.js`? (they look very close)

We use this in the FDC3 service so will need to put in the 4.x line.

PR #682, opened 2020-05-28
fix: #id 25683 Don't submit rename if newName is same as oldName

clausjs:
fix: #id [25683]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/25683/details)

**Description of change**
* When a user submits a rename of a folder the new text is checked against the old folder name. If the values are equal the rename is not submitted

**Description of testing**
1. Run Finsemble configured with Advanced App Launcher
1. Create a custom folder with any name
1. Rename the folder. In the input, type in the same name the folder had before the rename
1. [ ] Folder name returned when submitted and a "(1)" was not appended

### DIFF
--- a/src-built-in/components/advancedAppLauncher/src/components/FoldersList.jsx
+++ b/src-built-in/components/advancedAppLauncher/src/components/FoldersList.jsx
@@ -185,7 +185,7 @@ export default class FoldersList extends React.Component {
 		if (oldName.trim() === newName.trim()) nameChanged = false;
 
 		// Names must be unique, folders cant share same names
-		if (folders[newName] && nameChanged) {
+		if (folders[newName]) {
 			let repeatedFolderIndex = 0;
 			do {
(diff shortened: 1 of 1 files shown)


src-built-in/components/advancedAppLauncher/src/components/FoldersList.jsx
```diff
@@ -177,8 +178,14 @@ export default class FoldersList extends React.Component {
...
 		}
+
+		// Check if the submission is the same text as the old name.
+		// If false, renaming will be skipped
+		let nameChanged = true;
+		if (oldName.trim() === newName.trim()) nameChanged = false;
+
 		// Names must be unique, folders cant share same names
-		if (folders[newName]) {
+		if (folders[newName] && nameChanged) {
```
meredith-ciq, line 188: Is this check even needed now? The only way to have folders with numbers automatically added will be clicking new folder now.

PR #684, opened 2020-05-28
fix: #id 25678 Flash input when folder name is invalid. Fix folder reordering issues

clausjs:
fix: #id [25678]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/25678/details)

**Description of change**
* Changes how folder order is maintained
* Keeps focus on invalid name inputs and prevents user from editing other folders until conflicts are resolved
* Adds an animation to input to indicate error

**Description of testing**
1. Start Finsemble with Advanced App Launcher enabled
1. Add a folder
1. Rename the folder to some text that is invalid like "@@"
1. Try to take focus off the input/submit. The input bottom border should flash red to indicate invalid input
1. [ ] Unable to edit other folders/unable to stop editing a folder in error

### DIFF
--- a/src-built-in/components/advancedAppLauncher/src/components/FoldersList.jsx
+++ b/src-built-in/components/advancedAppLauncher/src/components/FoldersList.jsx
@@ -24,12 +24,13 @@ export default class FoldersList extends React.Component {
 		}
 		// Reference to ontainer element of folder list
 		this.listDiv = React.createRef();
-		this.renameFolder = this.renameFolder.bind(this)
-		this.changeFolderName = this.changeFolderName.bind(this)
-		this.onFoldersListUpdate = this.onFoldersListUpdate.bind(this)
-		this.keyPressed = this.keyPressed.bind(this)
-		this.deleteFolder = this.deleteFolder.bind(this)
-		this.onFocusRemove = this.onFocusRemove.bind(this)
+		this.renameFolder = this.renameFolder.bind(this);
+		this.changeFolderName = this.changeFolderName.bind(this);
+		this.cancelEdit = this.cancelEdit.bind(this);
+		this.onFoldersListUpdate = this.onFoldersListUpdate.bind(this);
+		this.keyPressed = this.keyPressed.bind(this);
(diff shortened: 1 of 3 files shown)


src-built-in/components/advancedAppLauncher/src/components/FoldersList.jsx
```diff
@@ -212,21 +243,41 @@ export default class FoldersList extends React.Component {
...
 		}
 
-		const canDelete = folder.canDelete;
-		const canEdit = folder.canEdit;
+		let buttons = null;
+		if (canEdit || canDelete) {
+			if (isEditing) {
+				buttons = (
+					<span className='folder-action-icons'>
+						{canDelete && <i id='confirm-edit' className='ff-check-mark-2' title='Accept Rename' onClick={this.renameFolder.bind(this, folderName)}></i>}
```
meredith-ciq, line 268: Shouldn't this line check for canEdit instead of canDelete?

clausjs: Yup, switched the buttons at the last second cause the mockup had them in reverse locations. Forgot to change the boolean checks

clausjs: Since `isEditing` cannot be true without `canEdit` being true I don't even think that check is necessary at all on that line

clausjs:
> What is the impact if this change is pulled into a system without the order field? I see you check if it exists, but will that leave new folders able to be reordered with each other and not the old folders? What kind of testing did you do around folder reordering?

Any folders without an order prop will appear above folders that have it since their order won't change (return 0). This will not effect which folders _can_ be reordered, simply where they are in the order. Even for previous state any folders that don't have an 'order' prop; adding a new folder, reordering folders, or deleting a folder will force the order props to be re-written.

The only time this should be an issue is if a user has folders in a certain order and chooses not to prefer previous state so their appFolders.list is overwritten. The order the folders appear in will then be the order they are key'ed in appFolders.folders on re-initialization. I'm not sure what I can do to work around that

PR #693, opened 2020-07-17
Aal new structure transform

clausjs:
fix: #id [27837]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/27837/details)

**Description of change**
* Checks for folder[folderName].apps instances that are still typeof === "object" and converts them to arrays.
* If a conversion takes place, a Central Logger warning is logged, and a user notification is alerted

**Description of testing**
1. Run Finsemble 4.2 with the advanced app launcher and add some folders with some apps in them
1. Upgrade to 4.3 without clearing state
1. [ ] The launcher should come up, and Finsemble should throw an alert warning that the data structure is incorrect
1. [ ] The warning will only show once unless the distributed store is set to reseed on every run, until the foundation is fixed, the warning will be received everytime

### DIFF
--- a/src-built-in/components/advancedAppLauncher/src/stores/StoreActions.js
+++ b/src-built-in/components/advancedAppLauncher/src/stores/StoreActions.js
@@ -95,6 +95,7 @@ function initialize(callback = Function.prototype) {
 		}
 
 		data.folders = store.values.appFolders.folders;
+		validateFolderDataStructure();
 		data.foldersList = folderList || Object.keys(store.values.appFolders.folders);
 		data.apps = Object.keys(appList).length > 0 ? appList : store.values.appDefinitions;
 		data.tags = store.values.activeLauncherTags;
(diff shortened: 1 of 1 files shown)


src-built-in/components/advancedAppLauncher/src/stores/StoreActions.js
```diff
@@ -144,6 +145,30 @@ function appInAppList(appName) {
...
+ * to the new structure (Array vs object)
+ */
+function validateFolderDataStructure() {
+	Object.keys(data.folders).map(folderName => {
+		const folder = data.folders[folderName];
+		if (!Array.isArray(folder.apps)) {
+			const warning = "Advanced App Launcher distributed store foundation appears to use incorrect format. Please ensure all folders[folderName].apps are arrays instead of objects";
+			// If the structure is wrong, notify the user in hopes that the foundation will be fixed
+			FSBL.Clients.Logger.warn(warning);
+			FSBL.UserNotification.alert("system", "ONCE-SINCE-STARTUP", "Distributed Store Type Mismatch", warning);
```
sgd2z, line 160: That notification needs better wording. Maybe - Application Launcher Persistent Store has data stored in deprecated format - or something like that.

clausjs:
> This looks good, but how does the user fix this? The notification will keep bugging them until they change the format, but is there any other way to do that besides deleting and re-adding their apps? Would it be better to make the change for them? (We would only want to do that if they don't plan to downgrade).

We do make the change for them, which is why they shouldn't see it more then once unless they have 'preferPreviousState' set to false, if they continue to re-seed with incorrect format, they will continue to get the message. 

So "how do they fix this?". 

Either prefer previous state, or change the config that keeps getting passed in at every run

meredith-ciq:
If we're updating the store and it's the config they need to change shouldn't the warning message say that instead of saying the persistent store has bad data?

PR #697, opened 2020-07-27
fix: #id 27392 Fix workspace menu issues regarding updates

clausjs:
fix: #id [27392]

[Kanban link](https://chartiq.kanbanize.com/ctrl_board/18/cards/27392/details)

**Description of change**
* Updates preferences components whenever the view changes to get most up-to-date workspace information
* Removes the constant workspace updates for every action slowing UI down
* Adds a private method to WorkspaceClient to retrieve a list of workspace names
* Adds a method to WorkspaceClient to retrieve the publish update constants to be used by a front end component to potentially filter out constant updates
* Switches name lists over to new method to retrieve workspace names

**Description of testing**
1. Open the preferences component after enabling import/export
1. Make changes to workspaces using the Preferences component/workspace menu and ensure changes are reflected in both places
1. [ ] Changes are reflected across all UIs expecting workspace changes

### DIFF
--- a/src-built-in/components/userPreferences/src/app.jsx
+++ b/src-built-in/components/userPreferences/src/app.jsx
@@ -31,6 +31,9 @@ class UserPreferences extends React.Component {
 		UserPreferencesStore.addListener({ field: "activeSection" }, this.setActiveSection);
 	}
 	setActiveSection(err, data) {
+		// Update workspaces every time the view changes to get up-to-date information
+		UserPreferencesActions.getWorkspaces();
+		
 		this.setState({
 			activeSection: data.value
 		});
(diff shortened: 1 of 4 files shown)


clausjs:
> This looks good. It seems like most of the other update reasons don't have any reason to to change the workspace list on a publish update call. The only one I'm uncertain about is empty workspace. That gets handled by calling WSM.addWorkspace and them sending a publish with the reason LOAD_DATA_RETRIEVED. This reason is now skipped but I don't think addWorkspace will trigger an update either. addWorkspace is used at a lower level than the function that triggers the router message for new_workspace. Can you verify an empty workspace gets updated in the user preferences list?

Looks like the update is caught

PR #714, opened 2020-09-17
fix: #id 29519 Linker sometimes shows up at wrong position

sgd2z:
https://cosaic.kanbanize.com/ctrl_board/106/cards/29519/details/

**Description of change**
* move fittodom call to after the window is shown

**Description of testing**
1. Arrange two windows like this: 
image
1. Alternately hit the linker button on both 50 times
1. Linker window always shows up on the correct window

### DIFF
--- a/src-built-in/components/linker/src/stores/linkerStore.js
+++ b/src-built-in/components/linker/src/stores/linkerStore.js
@@ -52,7 +52,13 @@ var LinkerStore = Object.assign({}, EventEmitter.prototype, {
 
 			self.setState(queryMessage.data);
 			// deal with the case of added linker channels
-			FSBL.Clients.WindowClient.fitToDOM();
+			const shownListener = () => {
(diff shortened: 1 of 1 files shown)


src-built-in/components/linker/src/stores/linkerStore.js
```diff
@@ -52,7 +52,13 @@ var LinkerStore = Object.assign({}, EventEmitter.prototype, {
 
 			self.setState(queryMessage.data);
 			// deal with the case of added linker channels
-			FSBL.Clients.WindowClient.fitToDOM();
+			const shownListener = () => {
+				finsembleWindow.removeEventListener("shown", shownListener);
+				console.log("launcherparams10", queryMessage.data);
```
sgd2z, line 57:
```suggestion
```

clausjs (approved):
Looks good, works much better  now

NinjaManatee (approved):

PR #717, opened 2020-10-08
Links update

MattSwartz:
Update links on master

### DIFF
--- a/configs/application/manifest-local.json
+++ b/configs/application/manifest-local.json
@@ -27,8 +27,8 @@
 	"splashScreenImage": "http://localhost:3375/assets/img/FinsembleSplash.svg",
 	"appAssets": [
 		{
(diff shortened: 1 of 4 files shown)


configs/application/manifest-local.json
```diff
@@ -27,8 +27,8 @@
 	"splashScreenImage": "http://localhost:3375/assets/img/FinsembleSplash.svg",
 	"appAssets": [
 		{
-			"src": "https://assets.finsemble.com/assimilation/dev/pr/assimilation.zip",
-			"version": "5.0.0.2",
+			"src": "https://assets.finsemble.com/assimilation/assimilation_5.0.1.0.zip",
```
d4hines, line 30: Is this correct? I don't know off hand.

MattSwartz: Yup this is from the A-team and should be patched to 5.0.1.0.

PR #787, opened 2023-02-07
Clone app work

gerard-ciq:


src/WindowTitleBar/CloneButton.tsx
```diff
@@ -0,0 +1,173 @@
...
+	const generateWindowName = (componentType):string => {
+		//Note: This will only work in localhost or a secure environment.
+		return `${componentType}-clone-${crypto.randomUUID()}`;
+	}
+
+	/**
+	 * simplified getComponentState and returning a promise
+	 * @param win
+	 */
+	const getAsyncComponentState = (fsblWindow:FinsembleWindow) => {
```
kriswest, line 156: When moving this into service, maybe just go fix these functions to return promises

src/WindowTitleBar/CloneButton.tsx
```diff
@@ -0,0 +1,173 @@
...
+		await FSBL.Clients.StorageClient.save({
+			topic,
+			key: `activeWorkspace${windowName}${windowName}`,
+			value: componentState
+		})
+	}
+
+	/**
+	 * Unused. But can be used to clone a window other than the current window
+	 * (potentially not correct url if remote clone)
```
kriswest, line 92: see other comments, just have to retreive options to get current URL

src/WindowTitleBar/CloneButton.tsx
```diff
@@ -0,0 +1,173 @@
...
+				addWindowToTabGroup(response.finWindow.name)
+			}
+		}
+	}
+
+	/**
+	 * Finds local window's tab collection and add's clone to it if it exists. Creates a new tab collection otherwise.
+	 * @param name
+	 */
+	const addWindowToTabGroup = async (name) => {
```
kriswest, line 48: pass in your finsembleWindow so this works for remote windows too

src/WindowTitleBar/CloneButton.tsx
```diff
@@ -0,0 +1,173 @@
...
+			return;
+		}
+
+		if(data) {
+			await FSBL.Clients.WindowClient.addToTabCollection({
+				tabCollectionId: data.id,
+				windowNames: [name],
+				promotedTab: name
+			});
+		}
```
kriswest, line 61: first if could be an if/else for readability - if this function was called you are either creating a group or adding to an existing group

src/WindowTitleBar/CloneButton.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+		fsblWindow.setComponentState({fields})
+	}
+
+	/**
+	 * Originally used to set component state before spawn but set it immediately after spawn - potential race condition.
+	 * @param windowName
+	 * @param componentState
+	 */
+	const prepareComponentState = async (windowName:string, componentState:any) => {
```
kriswest, line 81: Pure musing, but I wonder if they will ever need to mutate the state for the clone... That could be possible for an API call (same way as we do with workspaces, pass in a fn that does the mutation), but not for a built-in copy button. Doing anything about this would be premature without client feedback however.

src/WindowTitleBar/CloneButton.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+		if(!bounds && staggerPosition) {
+			// Can this always be assumed to be a number?
+			spawnParams.top = spawnParams.top ? spawnParams.top as number + staggerPosition : staggerPosition;
+			spawnParams.left = spawnParams.left ? spawnParams.left as number + staggerPosition : staggerPosition;
+		}
+
+		const {response} = await FSBL.Clients.AppsClient.spawn(component, {...spawnParams, addToWorkspace: true});
+		// Race condition
+		prepareComponentState(spawnParams.name as string,componentState)
```
kriswest, line 34: You are right that theres a race here which you'll probably lose sometimes. To resolve some small work is needed in the Launcher and Workspace service.


Workspace service wipes out `componentState` [here](https://github.com/Finsemble/finsemble-mono/blob/main/packages/core/src/platform/services/workspace/workspaceManager.ts/#L1455-L1471) and should take an optional argument to set it instead. The relevant responder is [here](https://github.com/Finsemble/finsemble-mono/blob/main/packages/core/src/platform/services/workspace/workspaceService.ts/#L1095), called from [here](https://github.com/Finsemble/finsemble-mono/blob/ebefe7bb901d6823b65706929d1ace7e7e762ab1/packages/core/src/platform/services/workspace/workspaceAdminAPI.ts#L13) and called from `spawnWithUpdateData` [here](https://github.com/Finsemble/finsemble-mono/blob/ebefe7bb901d6823b65706929d1ace7e7e762ab1/packages/core/src/platform/services/window/Launcher/launcher.ts#L3792).

You'll have to create a new [SpawnParam](https://github.com/Finsemble/finsemble-mono/blob/ebefe7bb901d6823b65706929d1ace7e7e762ab1/packages/core/src/platform/services/window/types.ts#L555) for component state, pass it down through the spawn functions  to here and then pass it onwards to WorkspaceService to set-up.

kriswest (commented):
Great first pass. You are right about the race and needing to integrate into AppsClient/Launcher to eliminate - I've laid out the path to doing so in comments.

The final implementation should include:
- API call on AppsClient/LauncherClient: I suggest `cloneWindow` to mirror showWindow.
  - optional windowIdentifier to clone, if not passed use local window
  - check if the window is a singleton before cloning - refuse to clone singletons and error out.
- Optional windowTitlebar UI button,pre-integrated into titlebar in the same style as the alwaysOnTop button
  - Config to enable per window or on all windows as per always onTop `finsemble.apps.apps[].hostManifests.Finsemble.foreign.components.Window Manager.alwaysOnTopIcon` or `finsemble.Window Manager.alwaysOnTopIcon`
    - Make sure our built-in UI components that use titlebars have the config set off by default so that notifications center, preferences, download manager, resolver and any others don't pick up the button when enabled for everything
    - I'll confirm final config for clone button with other PEs, in case they want to put under FSBLHeader (where always onTop should be really!).
    
Specifically, I propose: 
`AppsClient.cloneWindow(params: CloneParams, cb?): Promise<{ error?: StandardError ; response?: SpawnResponse }>`

where `CloneParams` is:
```ts
{
  /** if not set use local */
  windowIdentifier?: WindowIdentifier,
  /** add to exiting/create tab group with window being cloned */
  addToTabGroup=false: boolean,
  /** stagger window if not adding to tab group */
  staggerPixels=25: number, 
  /** override bounds (and ignore addToTabGroup and staggerPixels) */
  bounds?: WindowBounds
}
```

UI button will default to using:
`await AppsClient.cloneWindow({addToTabGroup : true})`

UI button to be enabled per window or on all windows as per alwaysOnTop `finsemble.apps.apps[].hostManifests.Finsemble.foreign.components.Window Manager.cloneWindowIcon` or `finsemble.Window Manager.cloneWindowIcon`.

Confirming that spec with other PEs here: 
https://finsemble.slack.com/archives/C02JAQ24CAE/p1675857332761709

src/WindowTitleBar/CloneButton.tsx
```diff
@@ -0,0 +1,173 @@
...
+			spawnParams = {...spawnParams, ...bounds}
+		}
+
+		if(!bounds && staggerPosition) {
+			// Can this always be assumed to be a number?
+			spawnParams.top = spawnParams.top ? spawnParams.top as number + staggerPosition : staggerPosition;
+			spawnParams.left = spawnParams.left ? spawnParams.left as number + staggerPosition : staggerPosition;
+		}
+
+		const {response} = await FSBL.Clients.AppsClient.spawn(component, {...spawnParams, addToWorkspace: true});
```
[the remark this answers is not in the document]
kriswest: Yes, in the options not the default config (as its set  by the launcher menu when spawned from there)

